PR #2, opened 2016-03-18
review

lukaselmer:


after_deployment_tasks.md
```diff
@@ -13,6 +13,14 @@ Check that you now
 
 * see "1+2=3" in each app.
 * have been redirected to https
+
+Setup cloudflare: https://www.cloudflare.com/cloudflare-settings?z=renuoapp.ch
```
diegosteiner: This direct link does not work for me :/ Maybe https://www.cloudflare.com/a/overview/renuoapp.ch ?

application_configuration.md
```diff
@@ -124,6 +111,7 @@ group :development do
   gem 'binding_of_caller'
   gem 'letter_opener'
   gem 'spring'
+  gem 'spring-commands-rspec'
```
diegosteiner: :+1:

application_configuration.md
```diff
@@ -209,6 +197,8 @@ git commit -v
 
 ## Devise
 
+Skip this if you will not use devise.lo
```
diegosteiner: What's the 'lo' for?

asset_delivery.md
```diff
@@ -1,6 +1,10 @@
 # Asset Delivery
 
-## Option A - Use Cloudfront CDN and Cloudflare CDN
+## Option A - Use Cloudfront Directly
+
+Since Cloudfront offers simple SSL certs, this now is an option.
```
diegosteiner: :+1:

google_analytics.md
```diff
@@ -34,4 +34,4 @@ ga('send', 'pageview');
 ```
 8. Write down the tracking ID (the string in the snippet with all caps and starting with an UX-XXX....) and note it - you will need it for the heroku config
 
-Do this for all 3 branches.
+Do this for all 3 branches, but instead of adding an account, add a property for the second and the third branch.
```
diegosteiner: :+1:

naming_conventions.md
```diff
@@ -4,16 +4,16 @@ The page lists obligatory rules for our naming conventions. Please stick to it.
 
 ## 1 Project Name
 
-The project name <project-name> consists of <base-project>-<sub-project>, and only uses [a-z0-9] and dash -. No underline _.
+The project name [project-name] consists of [base-project]-[sub-project], and only uses [a-z0-9] and dash -. No underline _.
```
diegosteiner: Maybe we should use it consistently throughout the book? E.g. in *after_deployment_tasks.md:20* there is still <application> used.

one_simple_test.md
```diff
@@ -54,6 +55,9 @@ class HomeController < ApplicationController
     render text: "1+2=#{val}"
   end
 end
+
+# sepc/spec_helper.rb
+SimpleCov.minimum_coverage 100
```
diegosteiner: :+1:

templates/bin/run
```diff
@@ -0,0 +1,6 @@
+#!/bin/zsh
+
+source ~/.zshrc
+
+(sleep 4 && open 'http://<appname>.dev:3000') &
+rails s -b <appname>.dev
```
diegosteiner: Maybe call ```bin/rails s``` to make use of spring?

after_deployment_tasks.md
```diff
@@ -13,6 +13,14 @@ Check that you now
...
+
+* Now open
+  * < application>-master.renuoapp.ch/home/check
+  * < application>-develop.renuoapp.ch/home/check
+  * < applicaiton>-testing.renuoapp.ch/home/check
```
schmijos: typo `applicaiton`

diegosteiner:
Only minor things found! Other than that: LGTM

application_configuration.md
```diff
@@ -106,12 +90,15 @@ gem 'devise-i18n'
 gem 'figaro'
 gem 'font-awesome-rails'
 gem 'goldiloader'
+gem 'http_accept_language'
+gem 'i18n-docs'
```
schmijos: We only need this dependency in *development* environment, right?

application_configuration.md
```diff
@@ -106,12 +90,15 @@ gem 'devise-i18n'
...
 gem 'jquery-rails'
 gem 'jquery-turbolinks'
 gem 'pg'
 gem 'rack-google-analytics'
 gem 'rails'
 gem 'rails-i18n'
+gem 'rails_real_favicon'
```
schmijos: It it really worth the dependencies (`rails < 5`, `rest-client ~> 1.8`), the few *github* stars, the extra config file and the extra [config line](https://github.com/RealFaviconGenerator/rails_real_favicon/blob/master/test/dummy/config/application.rb#L7)?

If it was, I would at least put it to the *development* environment section.

application_configuration.md
schmijos: :+1:

asset_delivery.md
schmijos: Are there any restrictions? For example should we limit *cloudfront* to deliver only the assets?

ci.md
```diff
@@ -2,34 +2,9 @@
...
-  - export DISPLAY=:99.0
-  - sh -e /etc/init.d/xvfb start
-  - sleep 3
-script: bin/check
-notifications:
-  email:
-    on_success: change
-    on_failure: always
-```
+If you need typescript, uncomment the tsd line.
```
schmijos: Which tsd line?

ci.md
```diff
@@ -2,34 +2,9 @@
...
-  - bin/setup
-  - export DISPLAY=:99.0
-  - sh -e /etc/init.d/xvfb start
-  - sleep 3
-script: bin/check
-notifications:
-  email:
-    on_success: change
-    on_failure: always
-```
```
schmijos: Where do I get the config from, now? Is it generated by the *renuo-cli*? If yes I would rename the title from `.travis.yml` to `Travis CI`

google_analytics.md
schmijos: :+1:

one_simple_test.md
```diff
@@ -34,7 +34,7 @@ It should be red (because of rspec).
 
 ```sh
 rails g controller home index check
-rm app/views/home/check.html.erb
+rm app/views/home/check.html.slim
```
schmijos: Is it clear that the generator being used is *slim*? What about:

    rm app/views/home/check.html.*

one_simple_test.md
schmijos: :+1:

one_simple_test.md
```diff
@@ -54,6 +55,9 @@ class HomeController < ApplicationController
     render text: "1+2=#{val}"
   end
 end
+
+# sepc/spec_helper.rb
```
schmijos: typo:

    # spec/spec_helper.rb

templates/.gitignore
```diff
@@ -11,3 +11,8 @@
 /Gemfile.local.rb
 /coverage
 .byebug_history
+/.env
+/bin/endless-rspec
+/app/assets/javascripts/**/*.js.map
+/app/assets/javascripts/**/*.js
+.tsdrc
```
schmijos: What about `/.idea/`?

templates/README.example.md
```diff
@@ -36,7 +36,7 @@ bin/check
 
 This runs
 
-* keyword check (console.log, TODO, puts, ...)
+* keyword check (console.log, T0D0, puts, ...)
```
schmijos: I think we don't like the todo checks. Talk with roger about that. He has good arguments.

templates/config/application.example.yml
```diff
@@ -14,7 +14,7 @@ DEVISE_SECRET_KEY: 'rake secret'
 DEVISE_PEPPER: 'rake secret'
 
 # Newrelic
-NEW_RELIC_LICENSE_KEY: 'TODO'
+NEW_RELIC_LICENSE_KEY: 'TO DO'
```
schmijos: oh please :-1:

after_deployment_tasks.md
lukaselmer: they just changed the url to https://www.cloudflare.com/a/dns/renuoapp.ch i guess. will fix it, thx!

application_configuration.md
lukaselmer: I don't know! :-o

lukaselmer: fixed it!

asset_delivery.md
lukaselmer: no, there are no restrictions (yet?). imo it's not needed, but let's discuss this face to face :) (actually, this doesn't concern this particular pull request - it's already the case with option b)

the difference from option c is that cloudflare is more difficult to configure and understand, and does not allow some configurations, e.g. timeouts (at least for me)

ci.md
lukaselmer: the one in the .travis.yml file

ci.md
```diff
@@ -38,10 +13,14 @@ The typings are loaded directly from GitHub. Unfortunately, GitHub has a rate li
 1. Go to https://github.com/settings/tokens and generate a new token without any permissions
 2. ```gem install travis```
 3. ```travis encrypt TSDRC_TOKEN=yourGeneratedToken --add```
-4. Add the following line to the before_script section: ```  - echo {\"token\":\"$TSDRC_TOKEN\"} > .tsdrc```
-5. ```echo ".tsdrc" >> .gitignore```
+4. Uncomment / add the following line to the before_script section: ```  - echo {\"token\":\"$TSDRC_TOKEN\"} > .tsdrc```
```
lukaselmer: @schmijos this tsd line ;)

after_deployment_tasks.md
lukaselmer: thx, fixed! but we need to use the same name everywhere in the future, so we can automatically replace them in all template files when copying (see also comment in the book where the template files should be copied)

application_configuration.md
lukaselmer: hmm, good call! yep, i think so!

lukaselmer: fixed!

application_configuration.md
lukaselmer: ah, i thought it needs to be in the production env. actually, i looked at the code here https://github.com/RealFaviconGenerator/rails_real_favicon/blob/master/lib/generators/favicon_generator.rb and it seemed really simple and also well tested https://github.com/RealFaviconGenerator/rails_real_favicon/tree/master/test

let's also discuss this face to face :+1:

naming_conventions.md
lukaselmer: yes, exactly my call (see also comment concerning the typo which @schmijos found).

the problem is that on gitbooks custom html "tags" ```<application>``` are invisible, since the browser doesn't know how to render them.

the next step would be to improve our cli, so that when we copy the template files we can replace the [project-name] placeholders with the real project name ;)

one_simple_test.md
lukaselmer: if you follow the guide, it will be slim (i know it, since this correction is because of that) :)

one_simple_test.md
lukaselmer: indeed! fixed! :D

templates/.gitignore
lukaselmer: it's actually already there, see line 5

lukaselmer: but yes, it's a good .idea

templates/.travis.yml
```diff
@@ -21,6 +21,7 @@ before_install:
 before_script:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
+# if you need tsd  - echo {\"token\":\"$TSDRC_TOKEN\"} > .tsdrc
```
lukaselmer: @schmijos and here is the commented tsd command

lukaselmer: btw: probably needs some work since we now should use https://github.com/typings/typings instead of tsd

templates/README.example.md
lukaselmer: yeah, let's talk about it face to face. it can be removed for certain projects if needed, but our definition of done includes "no todos in our code", so we should automate it if possible (actually, sometimes it also annoyed me [e.g. when i wanted to commit the readme file], but other times, it saved my commit)

templates/bin/run
lukaselmer: i think it should not make a difference? but yes, i think it's better to call bin/rails, fixing it!

templates/config/application.example.yml
lukaselmer: :cry:

ci.md
lukaselmer: you are right. when you do the whole setup guide, it's no problem, but if you want to add ci to an existing project, you are totally right!

lukaselmer: fixed

lukaselmer:
@diegosteiner @schmijos thank you very much for your valuable reviews! let's discuss the remaining stuff face to face at a :8ball:

templates/.gitignore
schmijos: sorry ^^

lukaselmer: Np xD

PR #6, opened 2016-04-19
Update .rubocop.yml

lukaselmer:
@coorasse what do you think?

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -15,6 +15,9 @@ Style/FileName:
   - 'Gemfile.local.rb'
   - 'Gemfile.local.example.rb'
 
+Style/FrozenStringLiteralComment:
+  Enabled: false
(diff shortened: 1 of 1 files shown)


coorasse:
I like this check. It prepares our apps for the new ruby version and I see no good reason in removing it.
It adds a line in all the files, right, this may be annoying, but it does with zero-pain for us.

lukaselmer:
> It adds a line in all the files, right, this may be annoying, but it does with zero-pain for us.
To me it feels painful to add an additional line of code to EVERY ruby file per project. Especially when Ruby 3 is far far away...? tbd with the team.

lukaselmer:
I'll revert the merge of the #5 until we've discussed this issue

diegosteiner:
TBD in a break (since it's easier than writing everything down).

PR #10, opened 2016-05-02
fix raven / sentry

lukaselmer:


after_deployment_tasks.md
```diff
@@ -55,7 +55,33 @@ TODO: update this since NewRelic has a newer way to do uptime monitoring.
...
+```
+
+## GetSentry Setup
+
+### JS Sentry Setup
+
+Below the raven.min.js, add the following to your application layout:
+
+```slim
+javascript: #{raven_js}
```
sihu: don't we have to check for the environment here?

PR #22, opened 2016-06-13
include css-comb

nicolaseckhart:
I don't think it adds any value to our code when the properties are sorted alphabetically and I don't feel like it's worth it spending time sorting them. Since there's no autocorrect I propose to disable this linter.


lukaselmer:
thanks for sharing! imo there are some advantages, e.g.

* no duplicated properties
* deterministic code / everyone writes it the same way

what do you think about something like https://github.com/csscomb/csscomb.js? i think this has automatic sorting.

ps: there are some auto formatters for the ide which are used by @cyrilkyburz @sihu @zorafuchs  https://plugins.jetbrains.com/plugin/7489?pr= https://plugins.jetbrains.com/plugin/7164?pr=

lukaselmer:
pps: https://github.com/csscomb/jetbrains-csscomb

nicolaseckhart:
while the first point about no duplicate properties is covered by other linters (eg [this one](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#duplicateproperty)) i do see your second point. 

something like cssbomb might be a good idea, because we we're discussing this last week and we feel it's not worth spending time manually sorting the css we write. 

while i won't be here next monday, i feel like it's worth bringing this small point up in a meeting (or maybe a break).

thanks for the response 👍

rbuetzer:
Additional advantage of sorting: Merging is much easier. I'm also pro sorting.

lukaselmer:
let's discuss this face to face :)
https://docs.google.com/document/d/16aX8XhIc7zqutNHp7TLqCXCZ-G1cy_A9AZley-3WnGU/edit#

templates/bin/check
```diff
@@ -60,6 +60,7 @@ if [ ! $? -eq 0 ]; then
   exit 1
 fi
 
+csscomb app/assets/stylesheets/**/*.scss
```
lukaselmer: @rbuetzer @nicolaseckhart this does not work! undo the merge?

nicolaseckhart: @rbuetzer why was this even merged?! it's still assigned to me?!?

nicolaseckhart: @lukaselmer thanks

PR #26, opened 2016-06-23
Feature/rm property sort order

lukaselmer:
@nicolaseckhart @rbuetzer i had to undo https://github.com/renuo/rails-application-setup-guide/pull/22 because it doesn't work like that. This PR replaces the old one.


templates/bin/check
```diff
@@ -60,6 +60,7 @@ if [ ! $? -eq 0 ]; then
   exit 1
 fi
 
+csscomb app/assets/stylesheets/**/*.scss
```
lukaselmer: @nicolaseckhart @rbuetzer this doesn't work like that, since the changes will be unstaged / not committed. Suggested workaround:
```git status``` before the command, ```git status``` after the command, and if it has changed, exit(1)

nicolaseckhart:
@lukaselmer how 'bout this?

lukaselmer:
@nicolaseckhart very nice, thx!

PR #29, opened 2016-07-14
Typings update

sihu:


typings.md
```diff
@@ -0,0 +1,17 @@
+# Typings / TS-Lint
+All typings-files are checked in. Make sure you don't have the `typings` in the `.gitignore`.
+
+typings.json:
+```json
+{
+  "name": "<<appname>>", // enter here your appname
```
lukaselmer: comments = invalid json

templates/tslint.json
```diff
@@ -1,15 +1,33 @@
 {
   "rules": {
+    "align": "todo",
```
lukaselmer: todos need to be fixed

typings.md
sihu: so you'd go for ` "name": "<<enter-here-your-appname>>",`

lukaselmer: aha! uhm, no, i would leave the comments away :)

lukaselmer: but i will offer you a better solution soon ;)

PR #30, opened 2016-07-27
Fix problems with CI

sihu:
we have problems with the ci, if we use <yourname> in the example. what happens is:

* ci copies the application.example.yml to  application.yml
* in the application.yml now we had an entry ADMIN_EMAIL: '<yourname>+sampleapp@renuo.ch' (not a valid mail)
* we can discuss if it makes sense to compare against ENV or if it should be mocked. but now tests comparing to `ENV['ADMIN_EMAIL']` will return a different mail. While locally it works, when have changed the `<yourname>`, on the CI it is not an array: 

```
       expected: ["<yourname>+sampleapp@example.com"]
            got: "<yourname>+sampleapp@example.com"
```

i would recommend to change it!

### DIFF
--- a/templates/config/application.example.yml
+++ b/templates/config/application.example.yml
@@ -6,7 +6,7 @@ APP_PORT: '3000'
 MAIL_HOST: ''
 MAIL_USERNAME: ''
 MAIL_PASSWORD: ''
-MAIL_SENDER: <yourname>+<application>@renuo.ch
+MAIL_SENDER: 'yourname+<application>@example.com' # change your name and domain to @renuo.ch
 
 # Secrets
 SECRET_KEY_BASE: 'rake secret'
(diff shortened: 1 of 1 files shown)


schmijos:
What i generally find problematic is that you cannot rely on `bin/setup` to create a working configuration for the app.

sihu:
also a nice suggestion of @coorasse was, to set the variable via `bin/setup` to `git config --global user.email`

sihu:
@schmijos i disagree with your change. the `<` and `>` where just there in the template for each project. i removed the ones which would stay in the application.example.yml. 
the ones which stay would have to be removed if you copy the example.yml (before you commit)

lukaselmer:
* It would be nice if we had a convention where / how the "renuo username" (e.g. lukas.elmer for lukas.elmer@renuo.ch) is stored (e.g. in an ENV variable). This way we could use bin/setup to replace the \<yourname> with the correct name automatically. Alternatively we can also ask the user for the yourname. And on the CI we just check if ENV['TRAVIS_CI'] is set, and if it does, we replace \<yourname> with travis.
  * PS: another idea was to create a `renuo clone` command to replace the placeholders...
* @sihu @schmijos I agree with @sihu. \<application> should be replaced when the project is setup, and should never be committed anyways. However, it would be nice if `renuo setup` would do that for us (created an issue for this here https://github.com/renuo/renuo-cli/issues/18)

What do you think?

sihu:
there was another idea to use the git email in the `git config`

PR #32, opened 2016-08-18
rubocop inheritance

lukaselmer:
inheritance from global rubocop config


sihu:
👍 nice one. i think we could discuss, what happens if the rubocop_base.yml changes. As far as i get it right now we could have new rubocop issues. But so far i also like the fact to be able to manage rubocop globally.

templates/.gitignore
```diff
@@ -16,3 +16,4 @@
 /app/assets/javascripts/**/*.js.map
 /app/assets/javascripts/**/*.js
 .tsdrc
+.rubocop-https---raw-githubusercontent-com-renuo-rails-application-setup-guide-master-templates--rubocop-base-yml
```
sihu: is this file generated? i couldn't find it!

lukaselmer: yep, it is as soon as rubocop is configured this way and runs for the first time

lukaselmer:
> As far as i get it right now we could have new rubocop issues (on CI) with a before working CI? But so far i also like the fact to be able to manage rubocop globally.

This is true. Especially if rubocop is upgraded. Having a global config file we can globally disable some rules here, so that we can upgrade rubocop, and the CI does not fail :)

But for sure, we have to be very careful when changing the rubocop file.

sihu:
as discussed. it totally makes sense and it should fail if changes procduce problems.

PR #33, opened 2016-08-23
Feature/csscomb additions

sihu:


templates/.csscomb.json
```diff
@@ -26,9 +26,9 @@
     "space-between-declarations": "\n",
     "strip-spaces": true,
     "unitless-zero": true,
-    "vendor-prefix-align": false,
+    "vendor-prefix-align": true,
```
sihu: linting told me that there must be true

templates/.csscomb.json
```diff
@@ -26,9 +26,9 @@
...
     "sort-order-fallback": "abc",
-    "tab-size": true,
+    "tab-size": 2,
```
sihu: linting told me that this must be a number:
```
Failed to configure "tab-size" option:
The option does not accept values of type "boolean".
Value's type must be one the following: number.
```

PR #39, opened 2016-08-31
add support folder to be required

cyrilkyburz:


lukaselmer:
Lgtm. I suppose it also works if there is no support dir.

cyrilkyburz:
Of course, it won't break the normal setup.

PR #41, opened 2016-10-19
Update .rubocop_base.yml

lukaselmer:
what do you think?


schmijos:
Are we still using *simple_form*?

lukaselmer:
@schmijos so can we merge it?

lukaselmer:
Yes, why not?

PR #43, opened 2016-10-27
avoid to override defaults

felixkubli:
https://github.com/bbatsov/rubocop/blob/master/manual/configuration.md#inheritance

says that 

> Configuration parameter that are hashes, for example PreferredMethods in Style/CollectionMethods are merged with the same parameter in the base configuration, while other parameter, such as AllCops / Include, are simply replaced by the local setting.

### DIFF
--- a/templates/.rubocop_base.yml
+++ b/templates/.rubocop_base.yml
@@ -13,6 +13,10 @@ Style/NonNilCheck:
 Metrics/BlockLength:
   Exclude:
   - config/initializers/simple_form_bootstrap.rb
+  - 'Rakefile'
+  - '**/*.rake'
+  - 'spec/**/*.rb'
+
 
 Style/FileName:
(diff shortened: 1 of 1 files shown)


lukaselmer:
> https://github.com/bbatsov/rubocop/blob/master/manual/configuration.md#inheritance
> says that ...

so?

templates/.rubocop_base.yml
```diff
@@ -13,6 +13,10 @@ Style/NonNilCheck:
 Metrics/BlockLength:
   Exclude:
   - config/initializers/simple_form_bootstrap.rb
+  - 'Rakefile'
+  - '**/*.rake'
+  - 'spec/**/*.rb'
+
```
lukaselmer: 1) additional line break is not necessary
2) why do we need these changes?

PR #46, opened 2017-05-02
Update .rubocop_base.yml

felixkubli:
https://github.com/bbatsov/rubocop/issues/4300

We need to add the `Gemfile`, as the default `Include` array gets overridden by ours


lukaselmer:
Thanks, lgtm in general. However, will this break any of our existing projects? I suggest that we try this config on some of them before applying it here.

coorasse:
Fixed in the new version

PR #50, opened 2017-07-10
Don't alter the underlying system with bin/setup

schmijos:


templates/config/application.example.yml
```diff
@@ -1,5 +1,5 @@
-# Use /etc/hosts to change the dns to 127.0.0.1, see bin/setup
-APP_HOST: '<application>.dev'
+# Edit /etc/hosts to let your DNS resolve to 127.0.0.1 or use puma-dev
+APP_HOST: '<application>.localhost'
```
schmijos: There will be a section about *puma-dev* in the laptop setup guide.
For the TLD change I suggest to read: https://iyware.com/dont-use-dev-for-development/

Kjir: I would read the comments as well and stick to `.dev`...

PR #54, opened 2017-08-15
Additions richi

rbuetzer:
This are some additions and clarifications I'd like to make after setting up Germann Auktionen.


gitignore.md
```diff
@@ -1,3 +0,0 @@
-# Add a .gitignore
-
-See https://raw.githubusercontent.com/renuo/rails-application-setup-guide/master/templates/.gitignore
```
rbuetzer: This section is not referenced in any chapter and the referenced file does not exist any more.

setup_heroku.md
```diff
@@ -8,4 +8,24 @@ to generate a script which will create all Heroku apps.
...
+If you think that the script is outdated please open a Pull Request.
+
+### Transferring Ownership of the application
+
+After creating a Heroku app, you have to transfer the ownership to `admin@renuo.ch`. This step may already be required
+to execute certain commands (like setting a `renuoapp.ch` subdomain) on the Heroku CLI.
+
+Transfer the ownership as follows:
+
+1. Login to [Heroku](https://id.heroku.com/login) and go to your app
```
coorasse: can we generate those commands in the script that creates the app?

setup_heroku.md
```diff
@@ -8,4 +8,24 @@ to generate a script which will create all Heroku apps.
...
+1. On the collaborator page, add the collaborator `admin@renuo.ch` ![](images/heroku/add_collaborator.png)
+1. On the settings page, transfer the ownership to `admin@renuo.ch` ![](images/heroku/transfer_ownership.png)
+1. If you do not have the credentials for `admin@renuo.ch`, ask wg-operations to accept the transfer of ownership
+
+
+### Creating a pipeline
+
+Once you have created *Heroku* apps for develop, master and testing, add them to a new pipeline with the same name as your
+application.
```
coorasse: is this possible in a script?

setup_heroku.md
rbuetzer: Yes, seems so: https://devcenter.heroku.com/articles/transferring-apps

However, the script needs to be split into two, resulting in the following workflow:

1. The first script creates the app and requests transfer. 
1. The Heroku admin accepts the transfer
1. The second script runs all commands which require a verified account

setup_heroku.md
rbuetzer: Yes, this is possible. I think I am going to do some investment on the Heroku setup script next week. Until then, I'd leave the description as is.

setup_heroku.md
coorasse: do we need a place where people can run those commands as a renuo admin? in the old setup you could connect to the server via SSH and run those command there. maybe we need something similar to be more efficient

rbuetzer: I think, we do not have to strive for maximum automatization here. Setting up a project happens 0-1 times per month. As long as there is always someone around who has the privileges to accept a transfer, I'd say this is fine.

According to https://xkcd.com/1205/, we can invest one or two hours to solve this, if we think, we will use it for the next five years.

coorasse: agree.

coorasse:
Please open a ticket for yourself to update the scripts. Thanks

PR #57, opened 2017-08-17
Update .rubocop.yml

coorasse:
this enables all the cops for the current rails version


schmijos:
Just be aware that this breaks some applications which use this template directly.

coorasse:
they shouldn't do it anymore :)

PR #58, opened 2017-08-17
Introduce a CI

coorasse:
This PR introduces a CI which runs a markdown linter and fixes all the issues so that is green


.mdlrc
```diff
@@ -0,0 +1,2 @@
+rules "~MD013", "~MD026", "~MD029", "~MD032", "~MD036"
```
coorasse: all disabled because have issues

PR #64, opened 2017-08-31
Bootstrap and SimpleForm

felixkubli:


ruby_on_rails/bootstrap.md
```diff
@@ -30,3 +36,52 @@ config.label_class = 'form-control-label'
 config.input_class = 'form-control'
 ```
 
+To make the error highlighting work you need to add some css to your application
+
+```scss
+.invalid-feedback {
```
coorasse: can you please explain me why this is necessary?

ruby_on_rails/bootstrap.md
```diff
@@ -30,3 +36,52 @@ config.label_class = 'form-control-label'
...
+    end
+  end
+end
+
+```
+
+Then you can add one to the input element:
+
+```ruby
+f.input :check, wrapper: :inline_checkbox
```
coorasse: you could also configure them as default wrapper so you don't have to specify it everytime, right?

ruby_on_rails/bootstrap.md
felixkubli: In the `4.0.0.beta` version of bootstrap, the `invalid-feedback` class has per default the attribute `display: none`. When the input element has the class `is-invalid`, this attribute gets overridden to `display: block`. Because I haven't found a way to add this class to the input element with the simple form configs, I override the class in the css.

ruby_on_rails/bootstrap.md
felixkubli: I didn't manage find out if that's possible.

ruby_on_rails/bootstrap.md
coorasse: is simpleform that adds `is-invalid`?

felixkubli: No. The Problem is that simpleform can't add this class.

ruby_on_rails/bootstrap.md
```diff
@@ -30,3 +36,52 @@ config.label_class = 'form-control-label'
...
+.invalid-feedback {
+  display: block;
+}
+
+.has-danger {
+  .form-control {
+    border-color: $form-feedback-invalid-color;
+  }
+}
+```
```
felixkubli: Specify that this is a workaround. Related simple form issue: https://github.com/plataformatec/simple_form/pull/1476

ruby_on_rails/bootstrap.md
felixkubli: `config.wrapper_mappings = { ... }`

PR #66, opened 2017-09-12
Update rubocop

coorasse:
Nice check! Unfortunately some gems depend on the order: i.e. a gem may require another gem to be loaded before, and if the "required" gem comes after the "requiring" one in alphabetical order, you are doomed 💣


schmijos:
Can you give an example, why it matters?

Bundler should resolve all dependencies itself correctly. I know of only one case when this can result in unexpected behaviour: If you mess up the [source priority](https://bundler.io/man/gemfile.5.html#SOURCE-PRIORITY). If your problem actually comes from having multiple sources, I suggest that we keep the linter and use source blocks like [this guy](https://www.codeoclock.com/2015/11/23/bundler-sources/)

Kjir:
The thing is: gems have to be ordered only inside the group. So this would give a warning:
```ruby
gem 'rails-rspec'
gem 'rspec'
gem 'bullet'
gem 'awesome_print'
```

But this would work:
```ruby
gem 'rails-rspec'
gem 'rspec'

gem 'awesome_print'
gem 'bullet'
```

coorasse:
Nice. Didn't know about it!

PR #67, opened 2017-09-12
Suggest to run db migrate to fix no schema.rb error

denismueller1:
when running bin/setup you get an error that there is no schema.rb. to fix this you have to run rails:db migrate. I updated the readme to explain this.


denismueller1:
Ok i fixed it & I will squash the pull request with a correct commit message anyways. Thanks!

coorasse:
Not sure: the schema.rb file **must** be there.
Maybe we need to add somewhere else in the setup that you have to run `bundle exec rails db:migrate` once to create the file (even if empty) or we change the ./bin/setup to take care of it.

PR #70, opened 2017-09-27
Change rails_helper template and correct the rspec guide

lukasbischof:
It may happen that SimpleCov doesn't recognise a class of the rails project when it has a name which is already used in an external gem (like 'Event'). 
In the SimpleCov guide they recommend that you should include SimpleCov at the top of the File (https://github.com/colszowka/simplecov#getting-started -> 2.), which solves this problem

### DIFF
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -45,7 +45,7 @@ Please check the [spec_helper template](../templates/spec/spec_helper.rb)
 
 * Inside `spec/rails_helper` we suggest to uncomment/enable the following:
 
-* before `require 'rspec/rails'`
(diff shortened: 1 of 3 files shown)


ruby_on_rails/rspec.md
```diff
@@ -27,6 +27,19 @@ you should know exactly why you are adding each one of them, why is necessary
...
 * delete the `test` folder
 
+* At the top of the `spec/spec_helper`
+
+```ruby
+require 'simplecov'
+SimpleCov.start 'rails' do
+add_filter do |source_file|
+source_file.lines.count < 5
+end
```
Kjir: Missing indentation

PR #84, opened 2017-12-14
Update .rubocop.yml

coorasse:
Disabled for two reasons:
1) It has many false positives including this (https://github.com/bbatsov/rubocop/issues/5223)
2) we don't use template strings


templates/.rubocop.yml
```diff
@@ -27,6 +27,9 @@ Naming/FileName:
 
 Style/FrozenStringLiteralComment:
   Enabled: false
+  
+Style/FormatStringToken:
+  Enabled: false
```
schmijos: Can you provide an example of why we need that?

PR #89, opened 2018-01-09
Improve SemaphoreCI setup

schmijos:
People apparently don't stick to the options configured on the image.


schmijos:
I somehow miss the famous line `ruby File.read '.ruby-version'`. Where would that be?

Kjir:
Semaphore doesn't support that (yet)

configure_ci.md
```diff
@@ -5,15 +5,26 @@ that all the tests pass before building and releasing a new version through our
...
-You should already be part of the Renuo organisation but you may not have permissions to add a project to the organisation.
-That's not a problem, you can configure it anyway and ask wg-operations to transfer it to the organisation afterwards (you have 100 free builds)
+You should already be part of the Renuo organisation but you may not have permissions to add a project to
+the organisation. That's not a problem, you can configure it anyway and ask wg-operations to transfer it to
+the organisation afterwards (you have 100 free builds)
 
-* Configure your CI to prepare and run your tests, depending on the application you are making the setup of.
+## Rails specific setup
 
-![semaphoreci_1](images/semaphoreci_1.png)
```
coorasse: please remove the image if is not necessary anymore.

coorasse:
@schmijos maybe it has been removed because Semaphore was not supporting it, so you won't find it in the guide. Since with your command that should work, feel free to also add that step.

PR #90, opened 2018-01-10
Add documentation for Cucumber

rbuetzer:


templates/features/env/database_cleaner.rb
```diff
@@ -0,0 +1,7 @@
+begin
+  DatabaseCleaner.strategy = :transaction
+rescue NameError
+  raise 'You need to add database_cleaner to your Gemfile (in the :test group) if you wish to use it.'
+end
+
+Cucumber::Rails::Database.javascript_strategy = :truncation
```
rbuetzer: By default, `rspec-rails` already puts this code into `features/env/custom_env.rb` as a comment. But I like having one file per concern instead of having one big custom config.

ruby_on_rails/cucumber.md
```diff
@@ -0,0 +1,86 @@
...
+The installation process of `cucumber-rails` is also documented on its
+[Github page](https://github.com/cucumber/cucumber-rails).
+
+To install `cucumber-rails`, add the following gems to your Gemfile:
+ 
+```rb
+group :test do
+    gem 'capybara', require: false
+    gem 'cucumber-rails', require: false
+    gem 'database-cleaner'
```
Kjir: I would make this future-ready: system tests and no database cleaner. Is it possible with cucumber to leverage that?

rbuetzer: I had a quick look, but didn't manage to get it running. But as it is would be nice to have, I'll invest another hour for that.

rbuetzer: Invested an hour. Did not find a way to integrate it. :-(

ruby_on_rails/cucumber.md
```diff
@@ -0,0 +1,86 @@
+# Cucumber
+
+[Cucumber](https://cucumber.io) is a testing framework which allows us to specify feature tests using plain language.
+In our projects, we use Cucumber for e2e tests. We integrate it using the
```
coorasse: I wouldn't state that since is not true. Also this is still the optional gems people can use. I would remove that sentence: "In our projects, we use Cucumber for e2e tests.". We also use rspec in some projects.

ruby_on_rails/cucumber.md
coorasse: cucumber does not support system tests yet therefore we need database cleaner. :(

ruby_on_rails/cucumber.md
```diff
@@ -0,0 +1,86 @@
...
+
+```rb
+group :test do
+    gem 'capybara', require: false
+    gem 'cucumber-rails', require: false
+    gem 'database-cleaner'
+end
+```
+
+Note that `capybara` and `database-cleaner` are optional. Most probably, you are going to need them.
```
coorasse: Why is that? Can you explain me when they would not be needed?

ruby_on_rails/cucumber.md
```diff
@@ -0,0 +1,86 @@
...
+
+```sh
+rails generate cucumber:install
+```
+
+At this point, it might be a good idea to inspect the generated files and commit. At least, ckeck if you run into any
+rubocop issues. Probably, you have to add an exception in `.rubocop.yml`:
+
+```yaml
+AllCops:
```
coorasse: too generic for me, also stating "probably" is not clear. Why is it causing rubocop issues? and which ones?

ruby_on_rails/cucumber.md
```diff
@@ -0,0 +1,86 @@
...
+  Exclude:
+    - 'lib/tasks/cucumber.rake'
+```
+
+## Running your first Cucumber test
+
+Add the following files:
+
+* [`app/controllers/home_controller.rb`](../templates/app/controllers/home_controller.rb) (If you haven't done so
+already) in the [RSpec](rspec.md) section.
```
coorasse: 👍

ruby_on_rails/cucumber.md
```diff
@@ -0,0 +1,86 @@
...
+
+```yaml
+AllCops:
+  Exclude:
+    - 'lib/tasks/cucumber.rake'
+```
+
+## Running your first Cucumber test
+
+Add the following files:
```
coorasse: I'd make two suggestions: one for new projects, where checking the home_controller.rb is a good idea. One for existing projects where I'd not add a controller just for that but I'd rather test an existing page.

ruby_on_rails/cucumber.md
```diff
@@ -0,0 +1,86 @@
...
+* [`features/home_check.feature`](../templates/features/home_check.feature)
+* [`features/step_definitions/home_check_steps.rb`](../templates/features/step_definitions/home_check_steps.rb)
+
+Now, you can run Cucumber in your terminal:
+
+```sh
+cucumber
+```
+
+You can also run specific Cucumber tests from RubyMine by hitting *Control* + *Shift* + *R*
```
coorasse: That's too much detail IMHO. That's not a rubymine guide 😄

ruby_on_rails/cucumber.md
rbuetzer: 👍 In our case (we use Cucumber for e2e tests), they are always needed. I am going to either remove this line or write down why we are using these gems.

ruby_on_rails/cucumber.md
rbuetzer: The file `.rubocop.yml` which is generated by `cucumber-rails` has lines which are too long. What do you think about rephrasing it as follows?

> At this time, it is recommended to inspect the generated files and commit. Before committing, make sure, Rubocop does not raise any issues. Therefore, you will have to add the following exception to `.rubocop.yml`:
> …

ruby_on_rails/cucumber.md
rbuetzer: In this case, I'd just add a sentence like

> If you add Cucumber to an existing project, test a real page instead of using the given example test.

ruby_on_rails/cucumber.md
coorasse: better. yes. I would prefer to fix the issues rather than adding an exception to rubocop. Which file is causing the issue? You wrote `.rubocop.yml` but I think you meant something different 😄

ruby_on_rails/cucumber.md
coorasse: perfect.

ruby_on_rails/cucumber.md
rbuetzer: `cucumber.rake` of course. I would not bother fixing the line length in this case, as the file is generated automatically by `cucumber-rails`.

coorasse: mmmm. I see your point but I disagree because we also fix rubocop issues from the files generated by Rails (for example). 
Apart from that if the line length is the only issue I'd rather propose to disable that specific cop (line length) for that file rather than `AllCops`

rbuetzer: I had another look, there are two cops failing: `Metrics/LineLength` and `Rails/FilePath`

Furthermore, he file in question starts with the following comment:

```
# IMPORTANT: This file is generated by cucumber-rails - edit at your own peril.
# It is recommended to regenerate this file in the future when you upgrade to a
# newer version of cucumber-rails. Consider adding your own code to a new file
# instead of editing this one. Cucumber will automatically load all features/**/*.rb
# files.
```

I would thus propose to treat it like an external library. This means, we don't write tests for it and we don't apply our coding guidelines. If we think there is something wrong with the file, we should open a PR on `cucumber-rails`.

PR #91, opened 2018-01-16
Update linter settings

mueden:


templates/.rubocop.yml
```diff
@@ -30,7 +30,11 @@ Style/FrozenStringLiteralComment:
   
 Style/FormatStringToken:
   Enabled: false  
-
+  
+Style/MixinUsage:
+  Exclude:
+    - 'spec/factories/**'
```
Kjir: Why?

templates/.reek
```diff
@@ -1,6 +1,8 @@
 ---
 Attribute:
   enabled: false
+NilCheck:
+  enabled: false
```
Kjir: Hmmmm... What's the reason for this?

templates/.rubocop.yml
mueden: if you have something like this

```ruby
include ActionDispatch::TestProcess

FactoryGirl.define do
  factory :company do
    sequence(:name) { |n| "Company #{n} AG" }
    logo { fixture_file_upload(Rails.root.join('spec', 'fixtures', 'test.png')) }
  end
end
```
it will give an error because the `include` is not within a module or a class

templates/.reek
mueden: It prevents usage of the safe navigation operator (`&.`).

templates/.rubocop.yml
```diff
@@ -30,7 +30,7 @@ Style/FrozenStringLiteralComment:
   
 Style/FormatStringToken:
   Enabled: false  
-
+
```
coorasse: this seems wrong to me...

PR #93, opened 2018-01-29
Add instructions on how to setup AWS S3

sihu:
To be discussed:

- [x] is s3.md at the right place in the `ruby`-folder, since we use it not just for ruby projects
- [ ] about cloudfront i'm not sure, if we should/must use it or if cloudflare can do the same. i had something in mind, that it's no longer needed.

### DIFF
--- a/ruby_on_rails/s3.md
+++ b/ruby_on_rails/s3.md
@@ -7,7 +7,7 @@ files
 
 ### Renuo-CLI is set up
 
-`gem install renuo-cli`
+`gem install renuo-cli` --> see [renuo-cli](https://github.com/renuo/renuo-cli)
 
 ### AWS-CLI is set up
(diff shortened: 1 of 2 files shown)


ruby_on_rails/s3.md
```diff
@@ -0,0 +1,60 @@
+# Amazon S3
+
+S3 is Amazon's Simple Cloud Storage Service‎, and used in most of your projects to store images and
+files
+
+## Preconditions
+
+### Renuo-CLI is set up
+
+`gem install renuo-cli`
```
coorasse: A link to the github repo can be useful

coorasse:
To me we need also the part to Cloudfront, yes. The file can also be moved outside the rails folder.

PR #94, opened 2018-02-05
introduce a pull request template

coorasse:


templates/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,29 @@
+### Description
+
+What I did (if it needs further explanation) and screenshots (if necessary).
```
schmijos: Shorten to "What I did and maybe screenshots of it"

templates/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,24 @@
+### Description
+
+What I did and maybe screenshots.
+
+### General
+
+- [ ] I reviewed my code.
+- [ ] This PR is relatable to a ticket number (via branch name, title or description).
+- [ ] I updated the ticket on Redmine.
+- [ ] *This PR is deployed on the testing environment*
```
schmijos: I'd leave this point away.

coorasse: Why? I think it's useful to know it has been deployed on testing for the reviewer.

schmijos: It's just not the normal case.

coorasse:
I'll wait for @rbuetzer as well to have an opinion also from the other team. We discussed that already in our team but maybe there are inputs also from them.

templates/PULL_REQUEST_TEMPLATE.md
macav: i'd say it can stay as optional point which i leave unticked if i didn't deploy it

PR #95, opened 2018-02-05
Add link to the page

schmijos:


js/README.md
```diff
@@ -0,0 +1,28 @@
+# Frontend apps
```
coorasse: It seems strange that the link states "Simple JS App" and the title of this page is a more generic "Frontend Apps"

js/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Trivial app
```
coorasse: I consider this section obvious, but probably it doesn't hurt to have it.

js/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+Consider using only static files (example: [Green Card Predictor](https://github.com/renuo/green-card-probability-frontend))
+and CDNs.
+
+## Simple JS app
+
+If you need a really simple JavaScript application and you need more than just an *index.html*, you can checkout this template:
+https://github.com/renuo/tamedia-altersheime/tree/0a5fc447bc7fdae9b81bbe0918ca64c864c58354
```
coorasse: 👍  I cannot go too much in detail in the template. I'd like if also @macav could take a look at it. 
I am just wondering if, for such a simple app, cucumber is not an overkill.. 🤔

js/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+The rest is done in these three files:
+
+* index.html
+* index.css
+* index.js
+
+## Complex JS app
+
+Use eighter React or [Angular](../angular/README.md).
```
coorasse: `either`

js/README.md
schmijos: I think cucumber is the easiest way to have tests at all. I think unit tests would be overkill.

js/README.md
schmijos: Yes.

PR #98, opened 2018-02-12
Improve AWS Setup Guide

sihu:


ruby_on_rails/aws.md
```diff
@@ -0,0 +1,60 @@
...
+
+You will need Renuo-CLI to be set up and at the newest version:
+
+`gem install renuo-cli` --> see [renuo-cli](https://github.com/renuo/renuo-cli)
+
+Make sure `renuo -v` shows the [newest version](https://github.com/renuo/renuo-cli/tags)
+
+### Start the Setup
+
+The following command will guide you through a setup to be able to run all the commands, you
```
coorasse: can you rephrase? I cannot understand the meaning of this sentence

sihu: what about 
```md
The following command will create commands to set up S3 and CloudFlare. Before that, the command makes sure, you have the aws-profile and the aws-cli set up, so you have the necessary rights to run the commands.
```

coorasse: better, now I got it :)

PR #99, opened 2018-03-05
Combine sparkpost sendgrid and mail setup to one file

denis-mueller:


ruby_on_rails/send_emails.md
```diff
@@ -33,19 +21,68 @@ class ApplicationMailer < ActionMailer::Base
...
+
+- Add the SendGrid AddOn to Heroku
+
+- add the following to `config/application.example.yml`
+
+```yml
+APP_DOMAIN: '[project-name].localhost'
+APP_PORT: '3000'
+MAIL_SENDER: 'yourname+<application>@example.com'
+SENDGRID_USERNAME: ''
```
coorasse: I don't like to have the specific service in the variable names. That's why we used a generic `MAIL_USERNAME` before

ruby_on_rails/send_emails.md
```diff
@@ -33,19 +21,68 @@ class ApplicationMailer < ActionMailer::Base
...
+SENDGRID_PASSWORD: ''
+```
+
+- Add this gem `sendgrid-ruby` to your gemfile and run `bundle`.
```
coorasse: Why do we need the gem?

ruby_on_rails/send_emails.md
```diff
@@ -33,19 +21,68 @@ class ApplicationMailer < ActionMailer::Base
...
+
+- Add the following code to your `config/environments/production.rb`
+
+```sh
+ActionMailer::Base.smtp_settings = {
+  user_name: ENV['SENDGRID_USERNAME'],
+  password: ENV['SENDGRID_PASSWORD'],
+  domain: ENV['APP_DOMAIN'],
+  address: 'smtp.sendgrid.net',
```
coorasse: if we use `MAIL_HOST` env variable we don't need to change this file

ruby_on_rails/send_emails.md
```diff
@@ -33,19 +21,68 @@ class ApplicationMailer < ActionMailer::Base
...
+  port: 587,
+  authentication: :plain,
+  enable_starttls_auto: true
+}
+```
+
+- Add the ENV variable `APP_DOMAIN` to the testing, develop and master project on heroku.
```
coorasse: isn't this variable already set in another step?

ruby_on_rails/send_emails.md
```diff
@@ -33,19 +21,68 @@ class ApplicationMailer < ActionMailer::Base
...
 
-* add the following to `config/environments/development.rb`:
+- add the following to `config/environments/development.rb`:
 
 ```ruby
 config.action_mailer.delivery_method = :letter_opener
 ```
 
-* add the following `config/environments/production.rb`:
+Now either setup SendGrid or SparkPost. If you need to send more than 100 mails per day use SparkPost otherwise use SendGrid. (It's more reliable and easier to set up.) (Sparkpost sometimes fails to send mails to outlook and bluewin email addresses.)
```
coorasse: mmmm. why did we use sendgrid the first time? should we just switch to it? I don't see sendgrid much more difficult to configure.

ruby_on_rails/send_emails.md
```diff
@@ -33,19 +21,68 @@ class ApplicationMailer < ActionMailer::Base
...
+  port: 587,
+  authentication: :plain,
+  enable_starttls_auto: true
+}
+```
+
+- Add the ENV variable `APP_DOMAIN` to the testing, develop and master project on heroku.
+
+- Add a Domain Whitelabel at <https://app.sendgrid.com/settings/whitelabel/domains>
+
```
coorasse: The part regarding the creation of an account on sendgrid is missing. which rules do we have to follow?

templates/.rubocop.yml
```diff
@@ -16,7 +16,7 @@ Naming/PredicateName:
 
 Metrics/BlockLength:
   Exclude:
-  - config/initializers/simple_form_bootstrap.rb
+  - 'config/**/*'
```
coorasse: this does not really belong to this PR, right?

ruby_on_rails/send_emails.md
mueden: But sendgrid automatically adds those to heroku so it would be super practical. Should I still rename them? (Then you would have to rename them 3 times on heroku, which costs time)

ruby_on_rails/send_emails.md
mueden: It is needed to access the sendgrid web api. The actual emails get sent from that api.

templates/.rubocop.yml
mueden: my bad :joy:

ruby_on_rails/send_emails.md
coorasse: I think this should be discussed in a wg-operations meeting. can you please add this point to the protocol, please?

ruby_on_rails/send_emails.md
coorasse: so we don't need them locally, am I right? We can just remove them...

ruby_on_rails/send_emails.md
coorasse: In the line 61 you configure the smtp server. and I don't see a SENGRID_API_KEY used anywhere. I think we don't need the gem with the actual configuration.

PR #101, opened 2018-04-04
Feature/improve bin scripts

lukasbischof:
### What I did

1. Updated the template files using `set -e` to automatically fail
1. Updated the guides and example snippets

### Why I did it

Currently, there are a lot of repetitive commands and echos that don't really add an advantage. Using `set -e` makes the files dryer and more readable

### DIFF
--- a/templates/bin/check
+++ b/templates/bin/check
@@ -1,28 +1,12 @@
 #!/bin/sh
 
+set -e
+
+bin/fastcheck
+
 if ! grep --exclude-dir="app/assets/typings/**" -i -r 'console.log' app spec
 then
   echo 'console.log found. Please fix them and try again, commit aborted'
-  exit 1
-fi
-
-if ! bundle exec mdl ./
-then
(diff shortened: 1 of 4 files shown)


templates/bin/check
```diff
@@ -1,28 +1,12 @@
...
-  bundle exec rubocop -a -D -c .rubocop.yml
-  echo 'Tried to auto correct the issues, but must be reviewed manually, commit aborted'
-  exit 1
-fi
-
-if ! bundle exec slim-lint app/views/ -c .slim-lint.yml
+if ! grep --exclude-dir="app/assets/typings/**" -i -r 'console.log' app spec
 then
-  echo 'slim-lint detected issues, commit aborted'
+  echo 'console.log found. Please fix them and try again, commit aborted'
```
coorasse: why did you keep this?

lukasbischof: Because the grep command does not output why it failed. Therefore, you would have no idea why the CI broke

PR #103, opened 2018-04-16
Configure linting to only be available in the development env

schmijos:


coorasse:
the CI should install only the gems required for the test environment, therefore also the linting ones. Makes sense to me, yes.

schmijos:
But now we don't tell the CI that it should only install test Gems, right?
image

coorasse:
You are right. That configuration you posted is wrong.
That's, for example, what we have in Hillary:
screen shot 2018-04-17 at 17 49 17
To me the linting, in our process, is strictly related to the tests: that's exactly why we do not let the tests pass if the linting fails. I don't see any reason why adding the gems to the test bundle should slow down the tests: we always have `require: false` therefore they are not required at runtime.

coorasse:
Since this is open since a week I'd add also Martin and Simon as reviewers and ask for further opinions.

macav:
i agree with @coorasse on this one

PR #104, opened 2018-04-23
remove reference to newrelic

macav:


ruby_on_rails/README.md
```diff
@@ -31,7 +31,6 @@ While everyone starts working there are some more things which you should setup.
 They are not optional, but the rest of the team can start working even if those are not in place yet.
 
 1. [GetSentry](getsentry.md)
-1. [NewRelic](newrelic.md)
```
coorasse: why? :)

macav: we don't handle the alerts and don't want to use it, so we agreed with Samuel to deprecate it. for now i disabled only the alerts, but i didn't close the account completely.

coorasse: to me, we need to have a tool to analyze potential performance issues in production. If we don't install newrelic by default anymore I'd like to have an alternative process.

macav:
let's keep it then

PR #106, opened 2018-04-25
Update .rubocop.yml

coorasse:
I think we should be allowed to perform direct updates on db in the tests


macav:
since you're on it, can you already update `TargetRailsVersion` to `5.2` ?

PR #108, opened 2018-05-25
Only use necessary rules for default setup

schmijos:
I think it's valuable to go back to the roots and reconsider from time to time if our rules are really of meaning. As I found out, they're not…

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -15,6 +15,9 @@ Style/MixinUsage:
   - 'bin/setup'
   - 'bin/update'
 
+Style/NonNilCheck:
+  IncludeSemanticChanges: true
+
 Metrics/BlockLength:
   Exclude:
   - 'spec/**/*.rb'
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -7,50 +7,14 @@ Metrics/LineLength:
 Style/Documentation:
   Enabled: false
 
-Style/NonNilCheck:
-  IncludeSemanticChanges: true
+Style/FrozenStringLiteralComment:
+  Enabled: false
 
-Naming/PredicateName:
+Style/MixinUsage:
   Exclude:
-    - 'app/serializers/**/*'
+  - 'bin/setup'
+  - 'bin/update'
 
 Metrics/BlockLength:
   Exclude:
-  - config/initializers/simple_form_bootstrap.rb
-  - 'Rakefile'
-  - '**/*.rake'
   - 'spec/**/*.rb'
-
-Naming/FileName:
-  Exclude:
-  - 'Gemfile'
-
-Style/FrozenStringLiteralComment:
-  Enabled: false
-  
-Style/FormatStringToken:
-  Enabled: false
-
-Rails/LexicallyScopedActionFilter:
-  Enabled: false
-
-Rails/SkipsModelValidations:
-  Exclude:
-    - 'spec/**/*.rb'
-
-AllCops:
-  TargetRailsVersion: 5.2
-  TargetRubyVersion: 2.5
-  Include:
-    - '**/Rakefile'
-    - '**/config.ru'
-  Exclude:
-    - 'bin/**/*'
-    - 'script/**/*'
-    - 'log/**/*'
-    - 'public/**/*'
-    - 'tmp/**/*'
-    - 'config/unicorn.rb'
-    - 'db/**/*'
-    - 'vendor/bundle/**/*'
-    - 'node_modules/**/*'


templates/.rubocop.yml
```diff
@@ -7,50 +7,14 @@ Metrics/LineLength:
...
 
-Rails/LexicallyScopedActionFilter:
-  Enabled: false
-
-Rails/SkipsModelValidations:
+Style/MixinUsage:
   Exclude:
-    - 'spec/**/*.rb'
+  - 'bin/setup'
+  - 'bin/update'
```
schmijos: With the rules until here, the standard Rails 5.2 setup is green (after autocorrect). Now follow our more custom rules.

templates/.rubocop.yml
```diff
@@ -7,50 +7,17 @@ Metrics/LineLength:
 Style/Documentation:
   Enabled: false
 
-Style/NonNilCheck:
-  IncludeSemanticChanges: true
```
schmijos: Kept, but moved down.

templates/.rubocop.yml
```diff
@@ -7,50 +7,17 @@ Metrics/LineLength:
...
-
-Naming/PredicateName:
-  Exclude:
-    - 'app/serializers/**/*'
```
schmijos: An app special case which is not generic.

templates/.rubocop.yml
```diff
@@ -7,50 +7,17 @@ Metrics/LineLength:
...
-
-Metrics/BlockLength:
-  Exclude:
-  - config/initializers/simple_form_bootstrap.rb
-  - 'Rakefile'
-  - '**/*.rake'
-  - 'spec/**/*.rb'
```
schmijos: I only kept rake tasks and specs (see below).

templates/.rubocop.yml
```diff
@@ -7,50 +7,17 @@ Metrics/LineLength:
...
-
-Naming/FileName:
-  Exclude:
-  - 'Gemfile'
```
schmijos: Not necessary anymore.

templates/.rubocop.yml
```diff
@@ -7,50 +7,17 @@ Metrics/LineLength:
...
-
 Style/FrozenStringLiteralComment:
   Enabled: false
-  
-Style/FormatStringToken:
-  Enabled: false
```
schmijos: I don't see the value, but we can also keep it if absolutely wanted.

templates/.rubocop.yml
```diff
@@ -7,50 +7,17 @@ Metrics/LineLength:
...
 
-Rails/LexicallyScopedActionFilter:
-  Enabled: false
```
schmijos: I think this rule makes a lot of sense.

templates/.rubocop.yml
```diff
@@ -7,50 +7,17 @@ Metrics/LineLength:
...
-Rails/SkipsModelValidations:
+Style/MixinUsage:
   Exclude:
-    - 'spec/**/*.rb'
+  - 'bin/setup'
+  - 'bin/update'
 
-AllCops:
-  TargetRailsVersion: 5.2
-  TargetRubyVersion: 2.5
```
schmijos: These are inferred by Rubocop automatically.

templates/.rubocop.yml
```diff
@@ -7,50 +7,17 @@ Metrics/LineLength:
...
 Style/FrozenStringLiteralComment:
   Enabled: false
-  
-Style/FormatStringToken:
-  Enabled: false
 
-Rails/LexicallyScopedActionFilter:
-  Enabled: false
-
-Rails/SkipsModelValidations:
```
schmijos: specs are excluded by default already.

templates/.rubocop.yml
```diff
@@ -7,50 +7,17 @@ Metrics/LineLength:
...
   Exclude:
-    - 'bin/**/*'
-    - 'script/**/*'
-    - 'log/**/*'
-    - 'public/**/*'
-    - 'tmp/**/*'
-    - 'config/unicorn.rb'
-    - 'db/**/*'
-    - 'vendor/bundle/**/*'
-    - 'node_modules/**/*'
```
schmijos: These folders are excluded by default already.

templates/.rubocop.yml
coorasse: this had false positives. that's why we decided to disable it. I am fine with re-enabling it and if there are still issues they should be addressed in the project itself.

templates/.rubocop.yml
coorasse: This rules brings you to write code like:
```
def show
end
```
which does nothing. since Rails doesn't require you to write ii if is empty. also more code = more bugs. in addition, to me could be considered dead code as well.

schmijos: Hm… right, you don't have to specify these actions in Rails. But do you like that? Should we really omit the one line `def show; end`? We're in MVC and I think it's just to much magic to omit the "C" of it (even more since we definitely have an "M" and a "V"). It's just that I personally find it very irritating, if there is **no** sign of an action (if you're lucky, there's a `before_filter :show` which indicated that something exists there).

PR #110, opened 2018-06-04
Bring in a Rails 5.1 setup

schmijos:
I recently set up a Rails 5.1 project. This is the setup of it.


templates/bin/setup
```diff
@@ -1,37 +1,37 @@
...
+  # Install JavaScript dependencies if using Yarn
+  # system('bin/yarn')
 
-  run 'Removing old logs and tempfiles' do
-    system 'rm -f log/*'
-    system 'rm -rf tmp/cache'
+  puts "\n== Copying sample files =="
+  unless File.exist?('config/application.yml')
+    system! 'cp config/application.example.yml config/application.yml'
   end
```
schmijos: This is custom for figaro. I myself use dotenv and this configuration:

    unless File.exist?('.env.local')
      cp '.env', '.env.local'
    end

templates/bin/setup
```diff
@@ -1,37 +1,37 @@
...
 
-  run 'Creating pre-push hook link', '.git/hooks/pre-push' do
-    system 'ln -s ../../bin/fastcheck .git/hooks/pre-push'
-  end
+  puts "\n== Preparing database =="
+  system! 'bin/rails db:reset'
```
schmijos: Here there's originally written `bin/rails db:setup`. `db:reset` drops it before, which makes sense in a local environment. Do you agree?

templates/bin/setup
```diff
@@ -1,37 +1,37 @@
 #!/usr/bin/env ruby
```
coorasse: To me it does not make a lot of sense to have this file here. We need to update it at every Rails change and seems a bit stupid. Why don't we remove it?

templates/bin/setup
coorasse: If the project has figaro, as we suggest in the guide, then this part makes sense. Maybe it should be moved in the figaro part of the guide. this also gives another reason to not keep this entire file here. what do you think?

templates/bin/setup
coorasse: doesn't matter. the schema drops the tables anyway.

templates/bin/setup
schmijos: Agreed.

templates/bin/setup
schmijos: Perfectly fine.

PR #113, opened 2018-06-13
Update linting_and_automatic_check.md

macav:


ruby_on_rails/linting_and_automatic_check.md
```diff
@@ -77,6 +77,26 @@ group :development, :test do
 end
 ```
 
+## ESLint
+
+> only if you'll use `es6` javascript.
+
+```
+yarn add eslint
+eslint --init (Use a popular style guide -> Airbnb)
```
macav: it is up for a discussion whether to use just `eslint:recommended`, or together with `google`, `airbnb` or `standard` presets. I picked airbnb because it has built-in rules applicable also for react and JSX and has some rules which google doesn't, such as preferring of object destructutring and disallowing underscored functions

coorasse: Fine for airbnb. Probably also @schmijos and @sihu want to give their opinion here

PR #124, opened 2018-08-07
Add robots.txt to our projects

coorasse:


ruby_on_rails/robots_txt.md
```diff
@@ -0,0 +1,40 @@
...
+```ruby
+# app/middleware/robots_txt.rb
+
+class RobotsTxt
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    if env['PATH_INFO'] == '/robots.txt' && ENV['BLOCK_ROBOTS'].present?
```
schmijos: I'd leave away the environment check because we only use this middleware if it is.

ruby_on_rails/robots_txt.md
```diff
@@ -0,0 +1,40 @@
...
+```ruby
+config.middleware.insert_before Rack::Sendfile, 'RobotsTxt' if ENV['BLOCK_ROBOTS'].present?
+```
+
+Add the variable `BLOCK_ROBOTS=true` in your develop and testing environment on Heroku:
+
+```sh
+heroku config:add BLOCK_ROBOTS=true --app [project-name]-testing
+heroku config:add BLOCK_ROBOTS=true --app [project-name]-develop
+```
```
schmijos: Would it complicate or clarify things if we'd add:

    heroku config:add BLOCK_ROBOTS=false --app [project-name]-master

?

coorasse: this would enable it anyway, because it checks just for `present?` maybe we should change to `ENV['BLOCK_ROBOTS']&.downcase == 'true'`

ruby_on_rails/robots_txt.md
coorasse: I don't really have a good point for that: I tend to duplicate this check in my projects and have it both in the middleware itself and in the project, but I am happy to leave this away.

ruby_on_rails/robots_txt.md
schmijos: No, `false.present?` evaluates to `false`

coorasse: yes, but makes sense, right? what would you prefer in the end?

schmijos: Leave it.

PR #126, opened 2018-08-27
Update reek configuration to reek 5

coorasse:


templates/.reek.yml
```diff
@@ -0,0 +1,40 @@
+---
```
coorasse: no option has been changed. they just got nested under the scopes `detectors` and `directories`

PR #128, opened 2018-08-27
Add rubocop-rspec to the setup guide

coorasse:
I prefer to have a strict check and disable/exclude/loosen the checks when we need that, rather than excluding a lot now.

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -1,3 +1,5 @@
+require: rubocop-rspec
+
 Rails:
   Enabled: true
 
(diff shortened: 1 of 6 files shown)


templates/.rubocop.yml
```diff
@@ -26,6 +28,21 @@ Metrics/BlockLength:
...
 
+RSpec/MultipleExpectations:
+  Exclude:
+    - 'spec/system/**/*'
+    - 'spec/requests/**/*'
+
+RSpec/MultipleExpectations:
+  Exclude:
+    - 'spec/system/**/*'
+    - 'spec/requests/**/*'
```
schmijos: According to my eyes, this is this is duplicate…

PR #131, opened 2018-08-27
Add rubocop-rspec to the setup guide

coorasse:


ruby_on_rails/rspec.md
```diff
@@ -58,30 +57,24 @@ Kernel.srand config.seed
...
 ```ruby
 require 'capybara/rspec'
 require 'capybara/rails'
 require 'selenium/webdriver'
 
-Capybara.register_driver :chrome do |app|
-  Capybara::Selenium::Driver.new(app, browser: :chrome)
-end
+RSpec.configure do |config|
+ # other configs
```
macav: is this comment needed?

macav: and is `config.use_transactional_fixtures = true` the default?

coorasse: right! I added it to the rails_helper.rb template

coorasse: for the comment: I wanted to show that the previous configurations are there, but is not in the template

templates/spec/rails_helper.rb
```diff
@@ -4,38 +4,23 @@
...
-      example.run
-    end
+  config.before(:each, type: :system, js: true) do
+    driven_by :selenium_chrome_headless
   end
-
-  config.include FactoryBot::Syntax::Methods
-
-  config.infer_spec_type_from_file_location!
 end
```
schmijos: It's kind of duplicated information. We have it in the rspec guide and here.

coorasse: Indeed, but I like to see, while following the setup guide, what I need to add, and at the same time a copy of the final version of the file as a reference.

PR #139, opened 2018-09-27
Follow Google recommendations

samuelsteiner:


google_analytics.md
```diff
@@ -52,3 +53,5 @@ config.middleware.use(Rack::Tracker) do
   end
 end
 ```
+
+Make sure you insert this script in the <head> of the page (not in the <body>)
```
coorasse: this configuration does not allow to decide where to place the script. The library rack tracker puts it for you already in the `head` https://github.com/railslove/rack-tracker#custom-handlers

PR #140, opened 2018-10-01
Ease ruling

sihu:


templates/.reek.yml
```diff
@@ -13,6 +13,16 @@ directories:
     "app/helpers":
       UtilityFunction:
         enabled: false
+      FeatureEnvy:
+        enabled: false
+    "app/jobs":
+      UtilityFunction:
+        enabled: false
+      InstanceVariableAssumption:
```
coorasse: the `InstanceVariablesAssumption` is a really helpful check. I'd like to discuss the removal of this one.

templates/.reek.yml
```diff
@@ -13,6 +13,16 @@ directories:
     "app/helpers":
       UtilityFunction:
         enabled: false
+      FeatureEnvy:
```
sihu: we think that helpers are meant exactly to extract some logic which could be also done in a model but don't belong there

templates/.reek.yml
```diff
@@ -28,6 +36,8 @@ directories:
     "app/mailers":
       InstanceVariableAssumption:
         enabled: false
+      FeatureEnvy:
```
sihu: it makes sense that in rails it's like that

templates/.reek.yml
```diff
@@ -13,6 +13,14 @@ directories:
     "app/helpers":
       UtilityFunction:
         enabled: false
+      FeatureEnvy:
+        enabled: false
+    "app/jobs":
+      UtilityFunction:
+        enabled: false
+    "app/services":
```
sihu: looked at some examples. it's a common pattern to use this `run` and there it would for example complain

templates/.reek.yml
```diff
@@ -13,6 +13,14 @@ directories:
     "app/helpers":
       UtilityFunction:
         enabled: false
+      FeatureEnvy:
+        enabled: false
+    "app/jobs":
+      UtilityFunction:
```
sihu: same as for service.

PR #141, opened 2018-10-09
bin/run should not run the browser as well

coorasse:
I think this script should be responsible of running your app.

### DIFF
--- a/templates/bin/run
+++ b/templates/bin/run
@@ -1,4 +1,3 @@
 #!/bin/sh
 
-(sleep 4 && open 'http://[project-name].localhost:3000') &
 bin/rails s


templates/bin/run
```diff
@@ -1,4 +1,3 @@
 #!/bin/sh
 
-(sleep 4 && open 'http://[project-name].localhost:3000') &
```
coorasse: why the sleep?

schmijos: Finally! I found this very disturbing. L introduced this for comfort. The sleep should give the browser time to open.

sihu: @coorasse @schmijos I guess the sleep would not give the browser time to open (becase it gets opened after 4s), but to the server to spin up. see it the same as you guys. everyone can add it to his/her personal setup if wanted (e.g. hotel-app)

PR #142, opened 2018-10-15
Fix for development

sihu:
I would recommend to not set this for `development` since then we have to add an URI to the `application.yml`. Other option would be to check for presence. There i see the risk of missing it on production.


sihu:
then i would go for `present?`. What do you think of it? Or we document a recommended value for the URI?

macav:
i'd go for `present?`

PR #146, opened 2018-11-10
Enable frozen-literal-string

coorasse:
This rubocop check has been disabled one/two years ago.
I remember that I had a discussion if we should have enabled it or not.
I'd like to bring back this topic and discuss the possibility of re-introducing it.
The main reasons to have this comment are:
* it will be the default in ruby3 (be ready)
* improves performances and memory usage for free

Some refs:
* https://freelancing-gods.com/2017/07/27/friendly-frozen-string-literals.html
* https://stackoverflow.com/questions/37799296/what-does-the-comment-frozen-string-literal-true-do
* https://bugs.ruby-lang.org/issues/11473

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -13,9 +13,6 @@ Metrics/LineLength:
 Style/Documentation:
   Enabled: false
 
-Style/FrozenStringLiteralComment:
-  Enabled: false
-
 Style/MixinUsage:
   Exclude:
   - 'bin/setup'


macav:
personally i find this nicer than having the comment in every file

`RUBYOPT="--enable-frozen-string-literal"`

but overall yes, why not

lukasbischof:
Frozen string literal [got added to bundler in version 1.12](https://github.com/bundler/bundler/blob/1-12-stable/CHANGELOG.md#1120pre1-2016-02-09), which was in 2016. So I'd say that probably most of the gems are updated by now 🙂.

sihu:
i'm on the same page as @macav that i don't like the code in every file. so if it will be default in ruby3, we will no longer need it, right? i'm asking myself if it's worth to dirty the git history on each project? But if you think, it's reasonable, i'm ok with it!

coorasse:
> i'm on the same page as @macav that i don't like the code in every file. so if it will be default in ruby3, we will no longer need it, right? i'm asking myself if it's worth to dirty the git history on each project? But if you think, it's reasonable, i'm ok with it!

I don't think the git history is a problem here: we apply a change and is even done automatically from  rubocop, i.e. no pain.
Why are you worried about the git history? Can you elaborate further? Anyway, for new files it should not be a problem

sihu:
i just don't like the fact, that we touch quite every file in every project by changing that. i know that's it's automated - even though for git-annotations it makes your life more difficult

coorasse:
@sihu from what I understand you would agree to keep it for new projects, right? 
Since the App setup guide is mainly for new projects I would suggest to change it. 
Existing projects can always not adopt it: every project has its custom rubocop configuration anyway. 
I also agree that old projects should not be **forced** to do it.

sihu:
👍 ok. like this I totally agree. like this it makes sense!

PR #148, opened 2018-11-12
Update Spec configuration

coorasse:
* Introduces global failures aggregation, in order to allow more expectations per test
* Introduces a configuration to catch JS errors in system tests
* Adds the binstub command when configuring rspec
* Configures RSpec before the linters

### DIFF
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -21,11 +21,12 @@ group :test do
 end
 ```
 
-You should know exactly why you are adding each one of them, why is necessary
+You should know exactly why you are adding each one of them and why is necessary.
 
 ## Configuration
 
 * Install rspec via `rails generate rspec:install`
+* Create a bin stub ith `bundle binstubs rspec-core`
 * delete the `test` folder
 * At the top of the `spec/spec_helper`
 
@@ -53,6 +54,10 @@ config.profile_examples = 5
 config.order = :random
 
 Kernel.srand config.seed
+
+config.define_derived_metadata do |meta|
+  meta[:aggregate_failures] = true
+end
 ```
 
 Please check the [spec_helper template](../templates/spec/spec_helper.rb)
@@ -78,6 +83,27 @@ end
 ```
 
 to use headless chrome for system tests.
+
+### Optional: catch javascript errors
+
+If you want to catch Javascript errors in your system tests, you can add the following to the `rails_helper.rb`:
+
+```ruby
+config.after(:each, type: :system, js: true) do
+  errors = page.driver.browser.manage.logs.get(:browser)
+  if errors.present?
+    aggregate_failures 'javascript errors' do
+      errors.each do |error|
+        expect(error.level).not_to eq('SEVERE'), error.message
+        next unless error.level == 'WARNING'
+
+        STDERR.puts 'WARN: javascript warning'
+        STDERR.puts error.message
+      end
+    end
+  end
+end
+```
 
 Please check the [rails_helper template](../templates/spec/rails_helper.rb).
 
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -26,7 +26,7 @@ You should know exactly why you are adding each one of them and why is necessary
 ## Configuration
 
 * Install rspec via `rails generate rspec:install`
-* Create a bin stub ith `bundle binstubs rspec-core`
+* Create a bin stub with `bundle binstubs rspec-core`
 * delete the `test` folder
 * At the top of the `spec/spec_helper`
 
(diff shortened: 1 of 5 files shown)


ruby_on_rails/rspec.md
```diff
@@ -21,11 +21,12 @@ group :test do
 end
 ```
 
-You should know exactly why you are adding each one of them, why is necessary
+You should know exactly why you are adding each one of them and why is necessary.
 
 ## Configuration
 
 * Install rspec via `rails generate rspec:install`
+* Create a bin stub ith `bundle binstubs rspec-core`
```
sihu: typo `with`

sihu: just to understand ist: this creates also a `bin/rspec`. is there other reasons to do it? https://bundler.io/man/bundle-binstubs.1.html

ruby_on_rails/rspec.md
```diff
@@ -53,6 +54,10 @@ config.profile_examples = 5
 config.order = :random
 
 Kernel.srand config.seed
+
+config.define_derived_metadata do |meta|
```
sihu: is there any known disadvantage in globally enabling failure aggregation? i could not find, that it makes it slower so i'm ok with it 👍

ruby_on_rails/rspec.md
```diff
@@ -79,6 +84,27 @@ end
 
 to use headless chrome for system tests.
 
+### Optional: catch javascript errors
+
+If you want to catch Javascript errors in your system tests, you can add the following to the `rails_helper.rb`:
+
+```ruby
+config.after(:each, type: :system, js: true) do
```
sihu: 🕺

templates/.rubocop.yml
```diff
@@ -29,9 +29,7 @@ Metrics/BlockLength:
   - 'spec/**/*.rb'
 
 RSpec/MultipleExpectations:
-  Exclude:
-    - 'spec/system/**/*'
-    - 'spec/requests/**/*'
+  Enabled: false
```
sihu: can you give me some background? i like the fact, that we try to follow that principle.

templates/spec/rails_helper.rb
```diff
@@ -23,4 +23,19 @@
   config.before(:each, type: :system, js: true) do
     driven_by :selenium_chrome_headless
   end
+
+  config.after(:each, type: :system, js: true) do
```
sihu: i'd consider to leave that piece of code and replace the one above  in ruby_on_rails/rspec.md with a link so we only have to update one.

templates/spec/spec_helper.rb
```diff
@@ -26,4 +26,8 @@
   config.order = :random
 
   Kernel.srand config.seed
+
```
sihu: same here: i'd go for one single source of truth

ruby_on_rails/rspec.md
sihu: by the way there is quite a nice article of a guy you might never heard of:
https://medium.com/@coorasse/catch-javascript-errors-in-your-system-tests-89c2fe6773b1

ruby_on_rails/rspec.md
coorasse:
```suggestion
* Create a bin stub with `bundle binstubs rspec-core`
```

coorasse: no. that's the main reason

ruby_on_rails/rspec.md
coorasse: not that I could find out. of course is "slower", meaning that it does not stop at the first error but it rather continue, therefore is "slower". but only in case of error, so it does not matter IMHO

templates/.rubocop.yml
coorasse: it goes together with the aggregate_failures. This cop checks that you have only one `expect` or use `aggregate_failures` when having multiple expectations. Since we enabled aggregate_failures globally, this check is not needed anymore.

ruby_on_rails/rspec.md
sihu: by convention we have the `bin/check` or `fastcheck` which does that. But maybe we want to use `bin/rspec` inside them?

coorasse: Yes, I would. I'll update these scripts.

coorasse:
I actually have problems implementing your suggestions, because it would mean to just point to the template, so either I remove the pieces of code from the instructions (and it would be difficult to understand what you have to add / write), or I remove the template which I don't find practical since is a very good reference. Any suggestion?

sihu:
Yeah, i see that point. guess there is no way to avoid it. Maybe we can improve the situation by adding a link underneath the code in the readme to the template? But i'm fine with it, also if you leave it like that 👍

PR #149, opened 2018-11-12
Disable RSpec/NestedGroups

sihu:
I do not see any advantage using this rule. IMO it does not lead to better code. i prefer rather to have more context but to be able to write well-structured specs.
https://www.rubydoc.info/gems/rubocop-rspec/1.7.0/RuboCop/Cop/RSpec/NestedGroups

I would avoid that writing test is to hard because of linters.

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -36,7 +36,7 @@ RSpec/ExampleLength:
     - 'spec/requests/**/*'
 
 RSpec/NestedGroups:
-  Enabled: false
+  MaxNesting: 5
 
 AllCops:
   Exclude:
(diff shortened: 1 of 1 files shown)


sihu:
I would not restrict it at all, since if you need 6 there might be a good reason for it. and i don't think that will be often the case. So i would just disable it.

sihu:
Had a short discussion with @schmijos. On one side this seems pretty random to set it to 5. On the other hand, if it's more than 5 it's not readable at all. I think complexity is anyways restricted due to complexity checks of rubocop as well. So it's kind of duplication. But I'm open to set it to 5 if you wish. `2` seems definitely to low.

coorasse:
I don't think any other cop would complain regarding complexity in rspec tests. I am in favour of `5`.

sihu:
No, not regarding complexity in rspec, but from the code: https://github.com/rubocop-hq/rubocop/blob/master/manual/cops_metrics.md#metricscyclomaticcomplexity

And i don't see a way to introduce more nested `context` than `if` which is in the end the Cyclomatic Complexity. Agree?

PR #156, opened 2018-11-27
Update .rubocop.yml

LionelLindt:
Configuration key `MaxNesting` for RSpec/NestedGroups is deprecated in favor of `Max`. Please use that instead.


coorasse:
Thanks!

PR #161, opened 2018-11-29
Allow sentry inline script

coorasse:
* By following the guide, we obtain CSP errors, since sentry is configured inline and this is forbidden by our CSP configuration.
* Calculating the nonce for each environment would be problematic because this is dynamic and might change if we change the script, but would be an alternative to this solution.
* At the moment this is the solution applied on auctionhub and bankentracker.
* This solution is better than having sentry disabled because of CSP therefore is an improvement, but also allows inline scripts.

Decision up to us, given that we need a solution.


macav:
whoops, i merged it by mistake already. if someone doesn't agree, feel free to revert it

PR #163, opened 2018-12-05
Add papertrail point

LionelLindt:


ruby_on_rails/README.md
```diff
@@ -42,6 +42,7 @@ and discuss with the PO how the service should be configured.
 1. [Uptime](uptime.md)
 1. Depending on the monitoring list, also Sentry notifications need to be configured.
 1. [Depfu security monitoring](depfu.md)
+1. [Papertrail monitoring](papertrail.md)
```
macav: I would make it clear that it should be enabled only for the projects based on the monitoring list, the same as Sentry

PR #166, opened 2019-01-29
Replace 1-star gem

schmijos:
I don't understand, how this could happen. `capybara-selenium` is a 1-star gem from 2014 which we introduced in August  2018.


schmijos:
The gems don't come by default. They work out-of-the-box but you need to have them in the _Gemfile_.

PR #177, opened 2019-04-04
Add Capybara and Selenium

schmijos:


ruby_on_rails/rspec.md
```diff
@@ -12,9 +12,11 @@ group :development, :test do
 end
 
 group :test do
+  gem 'capybara'
   gem 'factory_bot_rails'
   gem 'faker'
   gem 'shoulda-matchers'
+  gem 'selenium-webdriver'
```
coorasse: they are already provided by rails. does maybe `--skip-tests` influence this as well?

schmijos: Ah, this can be.

schmijos: Yes, it's the case. I'll undo the skip-test

PR #178, opened 2019-04-04
Exclude system tests from class describing RSpec

schmijos:
System tests should not be named based on module or classes, right?


coorasse:
I agree!

PR #179, opened 2019-04-04
Add Simplecov filters

schmijos:
File length of `< 5` is not enough (Rubocop frozen string literal comment). So let's get specific.


coorasse:
agree!

PR #185, opened 2019-04-09
Update bootstrap.md

LionelLindt:


ruby_on_rails/bootstrap.md
```diff
@@ -18,6 +18,7 @@ Add them with:
 and configure them in `environment.js`:
 
 ```js
+const webpack = require('webpack')
```
coorasse: why did you add this? a comment or description would halp me to review the PR.

LionelLindt: Doesn't the variable `webpack` that is used [here](https://github.com/renuo/applications-setup-guide/blob/22dea33199f9aa77e7f45a04fc437ac0f7aacdc5/ruby_on_rails/bootstrap.md#L22) first have to be required?

LionelLindt: image

PR #200, opened 2019-07-03
Refine Semaphore and Heroku setup

schmijos:
Please book your review time on 9999.

### DIFF
--- a/ruby_on_rails/configure_cd.md
+++ b/ruby_on_rails/configure_cd.md
@@ -1,23 +1,41 @@
 # Configure the Continuous Deployment (CD)
 
-You can configure continuous deployment on SemaphoreCI.
+We use Semaphore to deploy most of our apps. Some projects use CircleCI and
+mobile apps use Fastlane. This article is about how to deploy to Heroku with
+Semaphore.
 
-Configure an automatic deployment to each Heroku application (`[project-name]-master`,`[project-name]-develop`,
-`[project-name]-testing`). Use the API key of `admin@renuo.ch` for this task. wg-operations will help you, if you do
-not have access.
+## Prerequisites
 
-The deployment should execute the following deploy commands:
+* Heroku API key of `admin@renuo.ch` (*wg-operations* can help with that)
+
+## Setup `master` and `develop`
+
+Add two servers (very tiny plus button) and name them `[project-name]-master`
+and `[project-name]-develop`.
+
+Then edit the deploy commands to be the following:
 
 ```shell
 git push --force heroku $BRANCH_NAME:master
-heroku run -x "rails db:migrate || rails db:setup"
 ```
 
-this will deploy the code and migrate the database to the latest version.
+## Setup `testing`
 
-Name each server with the `[branch-name]` you are deploying.
+Add another server called `[project-name]-testing` and edit its deploy
+commands to be the following.
+
+```shell
+git push --force heroku $BRANCH_NAME:master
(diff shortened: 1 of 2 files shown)


ruby_on_rails/configure_cd.md
```diff
@@ -1,23 +1,41 @@
...
-`[project-name]-testing`). Use the API key of `admin@renuo.ch` for this task. wg-operations will help you, if you do
-not have access.
+## Prerequisites
 
-The deployment should execute the following deploy commands:
+* Heroku API key of `admin@renuo.ch` (*wg-operations* can help with that)
+
+## Setup `master` and `develop`
+
+Add two servers (very tiny plus button) and name them `[project-name]-master`
```
coorasse: I'd remove the explanation on how the button looks like.

ruby_on_rails/configure_cd.md
```diff
@@ -1,23 +1,41 @@
...
+and `[project-name]-develop`.
```
coorasse: I propose to name the servers only `master` and `develop`. the project name is already given by the current scope (you are in the project)

ruby_on_rails/configure_cd.md
```diff
@@ -1,23 +1,41 @@
...
+## Setup `testing`
+
+Add another server called `[project-name]-testing` and edit its deploy
+commands to be the following.
 
 ```shell
 git push --force heroku $BRANCH_NAME:master
-heroku run -x "rails db:migrate || rails db:setup"
+heroku pg:reset DATABASE_URL --confirm one11-testing
+heroku run -x "rails db:schema:load db:seed"
```
coorasse: I am not sure that we want to automatically destroy and rebuild the database at each deployment. What if I am testing some migrations? What if I need to keep data between two deploys?

ruby_on_rails/create_application_server.md
```diff
@@ -14,3 +17,14 @@ Please review the script before running it and execute only the commands you nee
...
 If you think that the script is outdated please open a
 Pull Request on the [renuo-cli](https://github.com/renuo/renuo-cli) project.
+
+## Setup Rails for Heroku
+
+Add a file called `Procfile` to your code root:
+
+```
+web: bundle exec puma -C config/puma.rb
+release: bundle exec rails db:migrate
```
coorasse: nice one 👍

ruby_on_rails/configure_cd.md
schmijos: This info would have helped me several times. The GUI is really bad in this regard.

ruby_on_rails/configure_cd.md
schmijos: Good point about the migrations. But then on the other hand: testing doesn't have a guaranteed state anyways. This change would enable you to make assumptions.

coorasse: why do you need to make assumptions on the state of the environment? If you want to reset it you can do it, you are free whatever you want on testing, but with this change, I won't have this freedom anymore.

schmijos: I don't agree with this point of view. I really expect the `testing` to be in a clean state after deployment. Actually it's exactly the same thing, I expect from `bin/setup` but just in production: **idempotence**.

But I'm not willing to waste time on this argument so I just follow you proposal.

coorasse: For Kinova we used it exactly for this reason: we mirrored production data and checked how the whole migration would have affected the production environment. This was great.

PR #202, opened 2019-08-19
Remove slim-lint

schmijos:
slim-lint doesn't autofix even the smallest issues which makes it a pain. So I suggest to remove it (or replace it with something that can autofix)

### DIFF
--- a/ruby_on_rails/linting_and_automatic_check.md
+++ b/ruby_on_rails/linting_and_automatic_check.md
@@ -61,16 +61,6 @@ To lint the SASS/SCSS files in our project we are going to use the `sass-lint` n
 Add to the project the linter configuration file you can find in the templates folder and check the `bin/fastcheck`
 template to see the command to execute the SCSS linting.
 
-## Slim lint
-
-> only if you'll use slim templates
-
-```ruby
-group :development, :test do
-  gem 'slim_lint', require: false
-end
-```
-
 ## ESLint
 
 ```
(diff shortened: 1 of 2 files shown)


schmijos:
Linters should make my life easier and not harder:

```
app/views/shared/_custom_flash_messages.html.slim:8 [W] LineLength: Line is too long. [122/120]
app/views/shared/_custom_flash_messages.html.slim:9 [W] TrailingBlankLines: Multiple empty lines in the end of file
app/views/shared/_default_flash_messages.html.slim:2 [W] RuboCop: Style/PercentLiteralDelimiters: `%w`-literals should be delimited by `[` and `]`.
app/views/shared/_default_flash_messages.html.slim:6 [W] TrailingBlankLines: Multiple empty lines in the end of file
```

Fo such a linter result I've to manually go to 4 different places in the code and correct one specific thing. That's not what I understand as an easy life.

coorasse:
In such a case, without slim-lint, you would have committed a 122 chars long line, and a file with multiple blank lines. Wouldn't this be bad?

sihu:
I am also more on the side of adding some autofixes for stupid findings

schmijos:
@coorasse I agree that it's bad. But I am questioning how bad it really is. Is a shiny golden slim template worth the fuzz?

coorasse:
some things are more important than others, a line longer than 120 characters is really annoying for example. 
we can, of course, discuss the single rules, but if the topic is a generic "yes/no to slim linting" I am for a yes: it helps me write better templates and keep them clean. 
If a reviewer would have noticed the above four things, commented, requested changes, etc... the fuzz would have been way bigger.

schmijos:
FYI: I started a thread here to feel the software engineering community outside of our bubble: https://softwareengineering.stackexchange.com/questions/396194/does-linting-without-auto-correction-make-sense

coorasse:
👍 for an opening to StackExchange. is a very good idea to get a different point of view.

PR #207, opened 2019-09-02
Allow merge commits

coorasse:
I'd like to be allowed to merge a PR without squashing for the following reasons:
* depending on the instructions given by the developer, the commits might make sense individually.
* if I am not given this option, as a reviewer I cannot choose what is best. 
* I'd expect commits that can be squashed to be squashed already when I review

### DIFF
--- a/configure_git_repository.md
+++ b/configure_git_repository.md
@@ -6,7 +6,6 @@ Please stick to it unless you have special needs.
 * Options
   * Features: Remove *Wikis*, *Issues* and *Projects*
   * Data services: Enable *Dependency graph* and *Vulnerability alerts*
-  * ~Merge button: Disable *Merge commits*~
 * Collaborators & teams
   * Add Renuo team as a collaborator with Admin access
   * Add Security team as collaborator with Write access


schmijos:
Probably hypothetical, but I'd like to mention my two reasons against merge commits:
* The commit context on the *develop* and *master* should always be a feature/pull request. There the CI is green. This assures that all commit in the whole history on develop and master are green. Like this you locally can git-bisect or git-blame in a very targeted way.
* People (me included) aren't taking care of their commits sometimes. These kinds of commits should not land in the master. Are **all** reviewers taking care of this issue? Could we lint at least the starting letter of the commit message being capital?

I'm fine if the reviewer makes an informed decision. My main problem is that squash merge is not the default.

@coorasse would these reasons make you reconsider?

coorasse:
Regarding the comments:
@sihu I care that commits are mainingful. When you need to go back in history to identify what changed and for which reason, is really important to have a proper commit history.
@schmijos I agree 100% with your first point. Regarding the second point there are two things:
* I think a good developer should care about the history and always care about what is merged into develop. That's why I think both the developer and the reviewer should put attention when they merge. Allowing all the options leaves more doors open. You are right: it also leaves the doors open to make mistakes, but I am willing to let everyone choose every time.
* first capital letter in commits: I am open for that! In the past I made a git hook which was taking care of this automatically. I think we could re-introduce it.

What do also the others think?

PR #208, opened 2019-09-27
Update configure_cd.md

simon-isler:


simon-isler:
> LGTM 👍 Make sure the CI does not fail

The Circle CI actually fails on master as well.

PR #209, opened 2019-10-09
new sentry configuration

coorasse:
I suggest a side-by-side review.

Main changes:
* use a npm package instead of our CDN, but still keep the CDN for the default configuration. (✅ done)
* use one project on sentry instead of three
* no configuration is needed for rails.
* use renuo-cli to automate configuration step (https://github.com/renuo/renuo-cli/pull/62)


ruby_on_rails/sentry.md
```diff
@@ -1,17 +1,23 @@
...
 
-* Add the `master` to the *#renuo* team, `testing` and `develop` to the *#no-notification* team. We are interested to receive notifications only regarding the `master` environment, but to log errors also in other environments.
+* Add the project to the *#renuo* team if the client pays for monitoring, to the `#no-notifications` otherwise.
 
-* Once you have created an entry, you will see the `dsn key` which you'll need in your config variables on Heroku. Note it.
+* Note the DSN key.
 
   ![sentry_dsn](../images/sentry.png)
 
+* Add the `SENTRY_DSN` value to the password manager.
```
macav: is it necessary? i always look it up on heroku or in sentry directly

coorasse: I had in mind of putting all the secrets you need for the `application.yml` in password manager, so you could find everything in one place instead of looking into different places.

sihu: mabye we could reformulate it then?
```suggestion
* Add the `SENTRY_DSN` value to the password manager as a copyable `application.yml`
```

schmijos: Even though I like the idea much, let's discuss pro and cons quickly.

coorasse: I remove it and we can discuss it separately

PR #213, opened 2019-10-17
Update sidekiq.md

sihu:
This is a very ugly one:

I had the problem that i updated my schedule.yml, but the task was still run. Now i found out why.

### DIFF
--- a/ruby_on_rails/sidekiq.md
+++ b/ruby_on_rails/sidekiq.md
@@ -76,7 +76,7 @@ bundle install
 schedule_file = 'config/schedule.yml'
 
 if File.exist?(schedule_file) && Sidekiq.server?
-  errors = Sidekiq::Cron::Job.load_from_hash(YAML.load_file(schedule_file))
+  errors = Sidekiq::Cron::Job.load_from_hash! YAML.load_file(schedule_file)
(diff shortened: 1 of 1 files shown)


ruby_on_rails/sidekiq.md
```diff
@@ -76,7 +76,7 @@ bundle install
 schedule_file = 'config/schedule.yml'
 
 if File.exist?(schedule_file) && Sidekiq.server?
-  errors = Sidekiq::Cron::Job.load_from_hash(YAML.load_file(schedule_file))
+  errors = Sidekiq::Cron::Job.load_from_hash! YAML.load_file(schedule_file)
```
sihu: 🐛 

* https://www.rubydoc.info/gems/sidekiq-cron/Sidekiq%2FCron%2FJob.load_from_hash
* https://www.rubydoc.info/gems/sidekiq-cron/Sidekiq%2FCron%2FJob.load_from_hash!

The one with the bang-operator removes the ones, that are not in the list (you can try it out with `Sidekiq::Cron::Job.all.map(&:name)`), while the current one just added the new ones.

schmijos: I'd encourage you to leave the parentheses stay.

coorasse: Agree.

PR #214, opened 2019-10-21
Use template token string format by default

macav:
```
# bad
format('%<greeting>s', greeting: 'Hello')
format('%s', 'Hello')

# good
format('%{greeting}', greeting: 'Hello')
```

The default in rubocop is the preferring `%<greeting>s`.

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -26,6 +26,9 @@ Metrics/BlockLength:
   - 'spec/**/*.rb'
   - 'config/environments/**/*.rb'
 
+Style/FormatStringToken:
+  EnforcedStyle: template
+
 RSpec/MultipleExpectations:
   Enabled: false
 


macav:
@coorasse what pushed me into opening this PR were problems with this rule on Schuler here:
https://github.com/renuo/schuler/blob/develop/config/routes.rb#L119

it complained about the template style, although this style is used also in the Rails documentation (https://api.rubyonrails.org/classes/ActionDispatch/Routing/Redirection.html) and although rubocop was happy when I changed it to the other format, it didn't work.

coorasse:
I see the issue, but I wouldn't make a rule for all the projects out of it. I would completely agree on an exception for this specific case/file.

macav:
ok, i'm closing this then, although i disagree with a default rule which doesn't work in some cases and is ugly too :)

PR #215, opened 2019-10-22
Avoid random failing tests

coorasse:
I fixed countless times random failing tests because this was missing. The last one was this morning. I'd like to have this as default.


coorasse:
To me is either/or. either we have it in `before` or in `after`. Having it in both doesn't seem necessary to me. I opted for the `after` because it feels like a "cleanup after yourself"

PR #220, opened 2019-11-05
Exchange command to start rails locally

schmijos:
The advantage: `tmp/pids` folder is created automatically.


coorasse:
do we need to prepend `bundle exec`?

macav:
for me it doesn't work. it works without `-u`: `rails s puma`. but since we only use `puma` web server by default, what is the difference between running simply `rails s`?

schmijos:
@macav is right. It starts puma automatically anyways.

PR #222, opened 2020-02-03
Enforced style

sihu:
According to this PR it's better to avoid slashes. I don't know why the default is different. Anyways i would like to not change all these paths in every project. Your opinion?

https://github.com/rubocop-hq/rubocop/pull/5571/files#diff-c9c43dc51d563f503df8494fc58ea077L507

image

### DIFF
--- a/ruby_on_rails/carrierwave.md
+++ b/ruby_on_rails/carrierwave.md
@@ -66,7 +66,7 @@ And also we want to clear the folder after the tests run in a `spec/support/carr
 ```rb
 RSpec.configure do |config|
   config.after(:suite) do
-    FileUtils.rm_rf(Dir[Rails.root.join("public#{UploaderBasepath::TEST_UPLOAD_PATH}")])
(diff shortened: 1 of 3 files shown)


sihu:
Ok. then i will change that PR and fix this also in other sources

ruby_on_rails/carrierwave.md
```diff
@@ -66,7 +66,7 @@ And also we want to clear the folder after the tests run in a `spec/support/carr
 ```rb
 RSpec.configure do |config|
   config.after(:suite) do
-    FileUtils.rm_rf(Dir[Rails.root.join('public', UploaderBasepath::TEST_UPLOAD_PATH)])
+    FileUtils.rm_rf(Dir[Rails.root.join("public#{UploaderBasepath::TEST_UPLOAD_PATH}")])
```
coorasse: this is fine like before. No need to change it

PR #227, opened 2020-03-26
Update instructions

sihu:


ruby_on_rails/sidekiq.md
```diff
@@ -4,7 +4,7 @@ As soon as you have to do work with background jobs please follow those suggesti
 They will help you having a proper system to do background work.
 We use [sidekiq](https://github.com/mperham/sidekiq) because it works well on Heroku and is easy to setup (suckerpunch for example causes memory issues on Heroku).
 
-* Add the following gem `gem 'sidekiq'` and install it
+* Add the following gem `gem 'sidekiq'` (to the `production` session only!) and install it
```
coorasse:
```suggestion
* Add the following gem `gem 'sidekiq'` to the `production` block and install it
```

ruby_on_rails/sidekiq.md
```diff
@@ -4,7 +4,7 @@ As soon as you have to do work with background jobs please follow those suggesti
 They will help you having a proper system to do background work.
 We use [sidekiq](https://github.com/mperham/sidekiq) because it works well on Heroku and is easy to setup (suckerpunch for example causes memory issues on Heroku).
 
-* Add the following gem `gem 'sidekiq'` and install it
+* Add the following gem `gem 'sidekiq'` to the `production` block and install it
```
sihu: After discussion with @macav i agree: we should distinguish: If you have the sidekiq-cron in place, you most likely want to use sidekiq also in development so you can test scheduled tasks. But if possible it should be only in `production`.

coorasse: Yes, that's true. Is also possibile that you use features of sidekiq not available in activejob and therefore you want to use it also in development and test. 
Still, this is a **setup** guide: you usually will do it like that, but of course you are free to change the configuration in the specific project to best fit the project's needs.

PR #228, opened 2020-03-30
Add -c flag to bin/fastcheck

kayzen-git:
This adds a -c flag to the fastcheck script on web to only check files that were actually changed. Everything still works as before without the flag for the CI.

image

### DIFF
--- a/templates/bin/fastcheck
+++ b/templates/bin/fastcheck
@@ -2,6 +2,34 @@
 
 set -e
 
+# Adds a -c flag to only check files that have actually changed
+# e.g.: bin/fastcheck -c
+CHANGED_ONLY="$1"
+changed() {
+  if [ "$CHANGED_ONLY" = '-c' ]
+  then
+    CHANGES="$(git diff --name-only | sed -e "s@^@$(git rev-parse --show-cdup)@g" | grep -E "$1" || :)"
+    if ! [ "$CHANGES" ]; then
+      echo "No changes found matching $1. Skipping." >&2
+      return 1
+    else
+      echo "Changed files:" >&2
+      echo "$CHANGES" >&2
+    fi
+  else
+    CHANGES=""
+  fi
+}
+
+# To use it wrap your command in an if statement like so:
+if changed "\.rb"
+then
+  # Command to be executed if -c flag is set and the regex matches any of the changed files OR if -c flag is not set.
+  # Use the $CHANGES variable to receive the list of changed files that matches the regex
+  # e.g.: bundle exec rubocop -D -c .rubocop.yml --fail-fast $CHANGES
+fi
+
+
 if ! bundle exec rubocop -D -c .rubocop.yml --fail-fast
 then
   echo 'rubocop detected issues!'


templates/bin/fastcheck
```diff
@@ -2,6 +2,34 @@
 
 set -e
 
+# Adds a -c flag to only check files that have actually changed
+# e.g.: bin/fastcheck -c
+CHANGED_ONLY="$1"
+changed() {
+  if [ "$CHANGED_ONLY" = '-c' ]
+  then
+    CHANGES="$(git diff --name-only | sed -e "s@^@$(git rev-parse --show-cdup)@g" | grep -E "$1" || :)"
```
coorasse: can you explain me what are you doing with this line? That basically boils down to the whole logic behind this PR.

coorasse:
What happens if the PR updates rubocop version? Only Gemfile.lock is changed but there might be new errors.

coorasse:
If the CI is still running the same way, why do we do that for rubocop? locally, it already does caching.

sihu:
I think especially for the linters it's really cool and much faster. Imagine slim lint etc. and for the rest (all the edge cases @coorasse mentioned, we still have the CI. I use the bin/check or bin/fastcheck locally and i think it's a nice feature, that you can just run all the specs of files you changed with a flag.

templates/bin/fastcheck
kayzen-git: The main flow goes something like this:
`git diff --name-only` - Get a list of changed filenames
`sed -e "s@^@$(git rev-parse --show-cdup)@g"` - replace the root path with the current workdir (`git rev-parse --show-cdup` returns the path from current dir all the way up to the git root) (the `@` in the sed command are an [alias](https://www.grymoire.com/Unix/Sed.html#uh-2) for `/` as we want to use paths inside our regex)
and `grep -E "$1"` just matches the files with the regex parameter $1

kayzen-git:
Yes Rubocop is able to do so already, but not all linters do that.

The main speed increase comes from the fact that it won't even start running the linters if the specified files haven't changed.

sihu:
> What happens if the PR updates rubocop version? Only Gemfile.lock is changed but there might be new errors.

We could even check for that in another PR: So in this case it would run it over all files again.

kayzen-git:
But yes, not everyone will use it, hence the opt-in flag.

coorasse:
I'd like to keep the ASG as small as possible. 
In our projects often these scripts differ a bit, for example in cobra we run system tests separated from all the others, and we have `check_linters` and `check_unit_tests` and other scripts. I think the ASG should be a starting point with everything you need, but then each project is completely free to diverge. 
I really like your proposal and I think it should be part of the ASG, but maybe not in the standard setup, also because it has corner-cases (I spent one minute thinking and I found one, are there others?), what about a separate chapter regarding this? On how to improve the speed of fastcheck?

coorasse:
I also like the idea. Problem is that at the moment I don't run it because is too slow. If the new version runs so fast that I don't notice then it will be super-cool to use it.

coorasse:
this PR is pending since more than 48 hours without further feedback. My suggestion is then to keep this valuable work and move it into a separate section of the ASG where we talk about "Here you will find a series of chapters and guides on how to setup some of the gems we use most often and some other useful services:"

coorasse:
closing for lack of feedback. Please feel free to re-open.

PR #230, opened 2020-04-20
Update .rubocop.yml

sihu:
These cops are new in rubocop and should be configured. Otherwise you get warnings in `bin/check`

To discuss: 

We already had some new cops of this kind. In version 1.0 they are supposed to be default. Should we then remove them again? If so, should we group and mark them somehow?

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -10,6 +10,15 @@ Rails/SkipsModelValidations:
 Layout/LineLength:
   Max: 120
 
+Layout/SpaceAroundMethodCallOperator:
+  Enabled: true
+  
+Lint/RaiseException:
+  Enabled: true
+
+Lint/StructNewOverride:
+  Enabled: true
+  
 Style/Documentation:
   Enabled: false
 
@@ -28,6 +37,9 @@ Style/HashTransformKeys:
   Enabled: true
 
 Style/HashTransformValues:
+  Enabled: true
+  
+Style/ExponentialNotation:
   Enabled: true
 
 Metrics/BlockLength:


sihu:
Maybe i have to tell you the background: Since some versions we get this annoying output for pending cops:
image

coorasse:
to add something to the discussion. rubocop was often bringing breaking changes with new cops introduced. That's why they don't enable them anymore by default. It's up to us, so we can upgrade rubocop and still not break the existing build. That's pretty cool.
Since this is an application setup guide, I'd expect here to have what is needed for a new setup, and therefore I agree to this change. This is also a guide to projects that want to upgrade to see what we decided to enable and what not. For the future, link  the docs to simplify the review process.
✅https://rubocop.readthedocs.io/en/latest/cops_layout/#layoutspacearoundmethodcalloperator

✅ https://rubocop.readthedocs.io/en/latest/cops_lint/#lintraiseexception

✅ https://rubocop.readthedocs.io/en/latest/cops_lint/#lintstructnewoverride

coorasse:
Since this was a topic in the last scrum meeting. I am nazi towards that. I use
```
AllCops:
  EnabledByDefault: true
```

in my personal projects. It takes me less time to fix or autofix new cops, then think about it. I disabled them if I don't like them afterwards.

schmijos:
> Since this was a topic in the last scrum meeting. I am nazi towards that. I use
> 
> ```
> AllCops:
>   EnabledByDefault: true
> ```
> 
> in my personal projects. It takes me less time to fix or autofix new cops, then think about it. I disabled them if I don't like them afterwards.

I'd prefer this solution against opting into new cops.

sihu:
Then let's go for that. Also fine for me!

coorasse:
closing in favour of 
```
AllCops:
  EnabledByDefault: true
```

PR #231, opened 2020-05-07
Remove lograge from suggested libraries

schmijos:
We often find ourselves debugging production environments with scarce because of that. I'd suggest that we don't suggest it as a default. We can still add it for heavy load apps.


schmijos:
@macav Good point. I like that suggestion.

PR #232, opened 2020-05-07
View spec exceptions

schmijos:
The default Rails view specs already exceed the default Rubocop limits of 5, e.g.:

```ruby
  it 'renders the edit answer_option form' do
    render

    assert_select 'form[action=?][method=?]', answer_option_path(@answer_option), 'post' do
      assert_select 'textarea[name=?]', 'answer_option[text]'
      assert_select 'input[name=?]', 'answer_option[veridical]'
      assert_select 'input[name=?]', 'answer_option[question_id]'
    end
  end
```

Also: I think it's ok to use instance variables in views specs since this is the Rails way (not in partials though).

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -50,6 +50,10 @@ RSpec/ExampleLength:
     - 'spec/requests/**/*'
     - 'spec/views/**/*'
 
+RSpec/InstanceVariable:
+  Exclude:
+    - 'spec/views/**/*'
+
 RSpec/NestedGroups:
   Max: 5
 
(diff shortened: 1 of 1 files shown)


schmijos:
So you'd use an instance variable like this?

    let(:answer_option) { @answer_option }

Isn't this a hack?

schmijos:
@coorasse I merged. Feel free to undo in a separate PR.

coorasse:
I don't really understand this PR. It states: 
"The default Rails view specs already exceed the default Rubocop limits of 5"
and it disables instance variables rubocop check. So I am not really sure what's that about 🤷‍♂️

PR #233, opened 2020-05-14
Reduce Heroku slug size

schmijos:
Webpacker produces 250mb JS files of no further use after asset compilation. Let's remove them.

### DIFF
--- a/ruby_on_rails/app_initialisation.md
+++ b/ruby_on_rails/app_initialisation.md
@@ -122,6 +122,17 @@ create the initializer for figaro in `config/initializers/figaro.rb`:
 * Enable the default Content Security Policies in `config/initializers/content_security_policy.rb`.
   The report URI will be set later in the step of Sentry configuration.
 
+* If you're using Webpacker, let's clean up after asset precompilation
+  to reduce Heroku slug size. Add this to the `Rakefile`:
+
+  ```ruby
+  Rake::Task['assets:clean'].enhance do
+    FileUtils.remove_dir('node_modules', true)
+    FileUtils.remove_dir('vendor/javascript', true)
+    FileUtils.remove_dir('tmp/cache/webpacker', true)
+  end
+  ```
+
 ## Finalising
 
 * Check if the following scripts run successfully: `bin/setup`, `bin/check`, `bin/run`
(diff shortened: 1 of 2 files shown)


ruby_on_rails/create_application_server.md
```diff
@@ -20,11 +20,21 @@ Pull Request on the [renuo-cli](https://github.com/renuo/renuo-cli) project.
...
+  It's read by Heroku to start the web app and worker jobs.
+
+* Add a file called `.slugignore` to your code root:
+
+  ```
+  /spec
+  /.semaphore
+  ```
+
+  Like this you can make files and folders to be excluded from the Heroku slug.
```
coorasse, line 39: love it

ruby_on_rails/app_initialisation.md
```diff
@@ -122,6 +122,17 @@ create the initializer for figaro in `config/initializers/figaro.rb`:
...
 
+* If you're using Webpacker, let's clean up after asset precompilation
+  to reduce Heroku slug size. Add this to the `Rakefile`:
+
+  ```ruby
+  Rake::Task['assets:clean'].enhance do
+    FileUtils.remove_dir('node_modules', true)
+    FileUtils.remove_dir('vendor/javascript', true)
+    FileUtils.remove_dir('tmp/cache/webpacker', true)
+  end
```
coorasse, line 133: love the idea. hate the name. calling it the same as the old `assets:clean`, that does a different job, causes me quite some confusion. what about `webpack:clean`? Also this would be confusing because the `assets:clean` task removes is responsible of removing old, compiled, assets, not in use anymore. this task does not clean the compiled assets. So...I like the idea. I'd name it differently, to avoid confusion.

kayzen-git:
Where will `assets:clean` be called? (As that's missing in the PR)
In the Procfile?

ruby_on_rails/app_initialisation.md
schmijos: This is not the full definition of `assets:clean` here. I use `enhance` to extend the existing task.

schmijos:
@macav @kayzen-git The assets are precompiled by Heroku just before the slug is created while `.slugignore` is applied even before asset compilation. This is why we need to keep `node_modules` around.

ruby_on_rails/app_initialisation.md
coorasse: Ah! Right! Then I am fine with it!

PR #235, opened 2020-05-24
Give some help to debug errors on the CI

coorasse:
When debugging errors on the CI, thanks to https://github.com/renuo/renuo-cli/pull/77 we can get the logs of the run. But how do we understand what the failing test wrote in the log? By default, there is no way to split the logs by test run. Thanks to this small information, we can understand that.


schmijos:
@sihu Feel free to revert.

PR #237, opened 2020-07-06
Update go_live.md

lukasbischof:
As a countermeasure to the forgotten db upgrade on Bio-Suisse, let's extend the db notice here


schmijos:
👍

PR #239, opened 2020-08-11
Changing gem name from database-cleaner => database_cleaner

Liberatys:
It seems like a typo has made its way into the repo 
Naming and Wiki:
https://github.com/DatabaseCleaner/database_cleaner

### DIFF
--- a/ruby_on_rails/cucumber.md
+++ b/ruby_on_rails/cucumber.md
@@ -14,7 +14,7 @@ To install `cucumber-rails`, add the following gems to your Gemfile:
 group :test do
     gem 'capybara', require: false
     gem 'cucumber-rails', require: false
-    gem 'database-cleaner'
+    gem 'database_cleaner'
 end
 ```
 


schmijos:
@Liberatys May I ask you to investigate if database cleaner is still needed? Because I think we had it only because of the Capybara tests, but we shouldn't need it anymore for Rails 5+.

coorasse:
That's for cucumber and is still needed: https://github.com/cucumber/cucumber-rails#installation

schmijos:
Hm… ok. Does someone know, why `cucumber-rails` doesn't run everything in a transaction?

coorasse:
Yes: https://github.com/cucumber/cucumber-rails/issues/406

schmijos:
@Liberatys @coorasse I made a ticket here, if someone want's to tackle this further: https://redmine.renuo.ch/issues/13938

Liberatys:
@coorasse @schmijos I appear to be late to the party :) 

@schmijos Is the ticket referring to replacing database_cleaner in the setup_guide or in general ?

Anyway :)

PR #242, opened 2020-08-31
Enforce no_braces for the last element of the array

coorasse:
Please check https://docs.rubocop.org/rubocop/cops_style.html#stylehashaslastarrayitem
I use the "no_braces" rule, as the opposite of the default "braces" because it removes useless syntax.


coorasse:
closing since this cops has bugs and weird behaviours like 
image

PR #243, opened 2020-08-31
Disable cop

coorasse:
https://docs.rubocop.org/rubocop/cops_style.html#stylehashaslastarrayitem
I don't like to add syntax sugar if not needed. The "no_braces" option doesn't work well if you have an array of hashes. see #242

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -20,7 +20,7 @@ Style/MixinUsage:
 
 Style/NonNilCheck:
   IncludeSemanticChanges: true
-  
+
 Style/HashEachMethods:
   Enabled: true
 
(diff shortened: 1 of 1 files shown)


macav (approved):

schmijos (changes requested):
@coorasse I agree with the issues you have with `no_braces`, but I'd be in favor of `braces` because I find everything else to be error-prone.

coorasse:
what do you mean with "error prone"? what kind of error can be made?

coorasse:
@sihu any opinion on the topic? Otherwise I will just go for a 2v1 win and merge it 😄

schmijos:
By error-prone, I mean that `[1, 2, one: 1, two: 2]` looks like it has 4 elements or that `[1, 2, one: 1, two: 2, 'three']` is invalid syntax. This cop set to `braces` helps us to make less errors.

sihu (approved):

coorasse:
I'll just go by majority. Thank you all 🙏

PR #244, opened 2020-08-31
Do not use syntax proposed by rubocop

coorasse:
I disagree with https://docs.rubocop.org/rubocop-rails/2.7/cops_rails.html#railswhereexists

Rails already removed the possibility of putting conditions in methods like `all` to force people to define conditions in `where`. I'd even disable the possibility to pass conditions to `exists?`

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -7,6 +7,9 @@ Rails/SkipsModelValidations:
   Exclude:
     - 'spec/**/*.rb'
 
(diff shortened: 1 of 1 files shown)


macav (approved):
I think it's not the same case as `all`, because `exists?` creates a different query `(SELECT COUNT(...) FROM ...)`. But I also don't have anything against disabling the cop

schmijos (approved):

PR #247, opened 2020-09-28
Update GitHub setup

Janis-Leuenberger:
Github played around with the settings so the guide was no longer up-to-date.

### DIFF
--- a/ruby_on_rails/README.md
+++ b/ruby_on_rails/README.md
@@ -25,12 +25,12 @@ It's now time to introduce some more tools which will help you and the team to k
 :tada: Finally you are ready to start working on you new project! :tada:
 
 While everyone starts working there are some more things which you should setup.
-They are not optional, but the rest of the team can start working even if those are not in place yet.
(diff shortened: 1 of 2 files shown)


schmijos (approved):

ruby_on_rails/README.md
```diff
@@ -25,12 +25,12 @@ It's now time to introduce some more tools which will help you and the team to k
 :tada: Finally you are ready to start working on you new project! :tada:
 
 While everyone starts working there are some more things which you should setup.
-They are not optional, but the rest of the team can start working even if those are not in place yet.
+Most are not optional, but the rest of the team can start working even if those are not in place yet.
 
 1. [Sentry](sentry.md)
-2. [NewRelic](newrelic.md)
+2. [NewRelic](newrelic.md) (optional)
```
coorasse, line 31: Since when NewRelic is optional?

schmijos: Isn't this something we install only for customers which pay for performance monitoring?

PR #249, opened 2020-10-28
Extract javascript error collection

Janis-Leuenberger:
Implemented this on my interns project and thought we might want to change here too.

### DIFF
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -112,21 +112,26 @@ Check that you see `1+2=3` in each app.
 
 ## Optional: catch javascript errors
 
-If you want to catch Javascript errors in your system tests, you can add the following to the `rails_helper.rb`:
+If you want to catch Javascript errors in your system tests, you can create the `support/javascript_error_collector.rb`:
 
 ```ruby
-config.after(:each, type: :system, js: true) do
-  errors = page.driver.browser.manage.logs.get(:browser)
-  if errors.present?
+module JavaScriptErrorCollector
+  def collect_errors
+    errors = page.driver.browser.manage.logs.get(:browser)
+
     aggregate_failures 'javascript errors' do
       errors.each do |error|
         expect(error.level).not_to eq('SEVERE'), error.message
-        next unless error.level == 'WARNING'
-
-        warn 'WARN: javascript warning'
-        warn error.message
+        expect(error.level).not_to eq('WARNING'), error.message
       end
     end
   end
 end
+```
+
+And then include it in the `rails_helper.rb`:
+
+```ruby
+config.include JavaScriptErrorCollector
+config.after(:each, type: :system, js: true) { collect_errors }
 ```
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -124,16 +124,31 @@ module JavaScriptErrorCollector
         expect(error.level).not_to eq('SEVERE'), error.message
 
         next unless error.level == 'WARNING'
-        puts "\e[33m\nJAVASCRIPT WARNING\n#{error.message}\e[0m"
+        warn "\e[33m\nJAVASCRIPT WARNING\n#{error.message}\e[0m" if Rails.application.config.show_javascript_warning
+
+        next unless Rails.application.config.fail_on_javascript_warning
+        expect(error.level).not_to eq('WARNING'), error.message
       end
     end
   end
 end
 ```
 
-And then include it in the `rails_helper.rb`:
+Then include it in the `rails_helper.rb`:
 
 ```ruby
 config.include JavaScriptErrorCollector, type: :system
 config.after(:each, type: :system, js: true) { collect_errors }
+```
+
+And configure `config/environments/test.rb` to your preference
+
+```ruby
+Rails.application.configure do
+  # Make the test fail when a warning is logged
+  config.fail_on_javascript_warning = false
+
+  # Ouput the warning into the console
+  config.show_javascript_warning = false
+ebd
 ```
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -115,5 +115,3 @@ Check that you see `1+2=3` in each app.
 * Create the module [`spec/support/javascript_error_reporter.rb`](../templates/spec/support/javascript_error_reporter.rb`)
 
 * Verify that `config.include JavaScriptErrorReporter, type: :system, js: true` is in your [`rails_helper.rb`](../templates/spec/rails_helper.rb)
-
-* Add `js: true` to any system test that should fail when JavaScript errors occur.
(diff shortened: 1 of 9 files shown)


ruby_on_rails/rspec.md
```diff
@@ -112,21 +112,26 @@ Check that you see `1+2=3` in each app.
...
       end
     end
   end
 end
 ```
+
+And then include it in the `rails_helper.rb`:
+
+```ruby
+config.include JavaScriptErrorCollector
```
coorasse, line 135: we don't need to include this always, right? shall we include it only in system tests?

ruby_on_rails/rspec.md
```diff
@@ -112,21 +112,26 @@ Check that you see `1+2=3` in each app.
...
+    errors = page.driver.browser.manage.logs.get(:browser)
+
     aggregate_failures 'javascript errors' do
       errors.each do |error|
         expect(error.level).not_to eq('SEVERE'), error.message
-        next unless error.level == 'WARNING'
-
-        warn 'WARN: javascript warning'
-        warn error.message
+        expect(error.level).not_to eq('WARNING'), error.message
```
coorasse, line 125: It seems to me that with this change the tests will fail also for warnings, while before they were failing for errors and just print a message for warnings. correct?

coorasse (changes requested):
If my assumption is correct then I disagree. I don't want to have failing tests because of warnings. If I am wrong then I need a bit more explanation regarding the PR :)

ruby_on_rails/rspec.md
Janis-Leuenberger: Yes this is true. In my opinion the warnings were lost in all the console output. And javascript warnings can be important so I therefore made the tests fail. Proposal: Don't let the tests fail but make the warnings more apparent. With colors or something.

schmijos: I think we should either fail on warnings or show nothing. Warnings tend to pollute logs without getting fixed.

I'm generally on the page that warnings were in control of should make the tests fail. **BUT:** are we really in control of the warnings? What I would not want, is that tests start failing because of a Chromedriver update.

Janis-Leuenberger: What if we let the tests fail if the warning comes from one of the projects scripts or the browser and ignore it if it comes from a dependancy?

schmijos: I generally like the idea.

But:

* Can you even do that? How clear is the source of a warning in JS?
* The problem with Chrome updates remains: if Chrome introduces new warning after an update, our builds will suddenly fail.

Janis-Leuenberger: It's possible but probably difficult to account for all possibilities/sources.

At this point I think it's best to go back to just print them. Is everyone ok with that?
@coorasse @schmijos

Janis-Leuenberger:
Example output:
Screenshot 2020-11-03 at 14 36 43

ruby_on_rails/rspec.md
```diff
@@ -122,7 +122,9 @@ module JavaScriptErrorCollector
     aggregate_failures 'javascript errors' do
       errors.each do |error|
         expect(error.level).not_to eq('SEVERE'), error.message
-        expect(error.level).not_to eq('WARNING'), error.message
+
+        next unless error.level == 'WARNING'
+        puts "\e[33m\nJAVASCRIPT WARNING\n#{error.message}\e[0m"
```
schmijos, line 127: I'd not show those in the console. This would pollute our nice green dots.

I'd leave this away except someone opts-in with a config setting.

Janis-Leuenberger:
Now you have the following options:
* Only fail on errors (default)
* Just fail on warnings
* Just log the warnings
* Do both

coorasse (changes requested):
Sorry but I am not a big fan of more options: they make the configuration more complex just to please everybody and in the end they just lead to more discussions. 

The application setup guide is not the bible anyway: if the team working on a project wants to implement the "catching of javascript errors" they can do it already as they please, the ASG provides just an example. 

One more thing is that these new options should probably be on rspec or capybara level, rather than rails itself.

My proposal: keep it as it is and feel free to adapt it in each single project. 

But also: we are a large team, some others might have different opinion and the majority wins 😄

schmijos:
@coorasse I think he talks about that this config is optional in the ASG.
@Janis-Leuenberger I don't understand why you should configure anything in the Rails config.

My 3 cents on the current proposal:

1. Include the JS error collector per default into the Renuo RSpec configuration (not optional, like it's written now)
2. Configure it to collect errors for all tests flagged with `:js` and fail if there are any.
3. Drop what ever you started with the `Rails.config`

@coorasse Agreed?

ruby_on_rails/rspec.md
```diff
@@ -112,21 +112,43 @@ Check that you see `1+2=3` in each app.
...
     end
   end
 end
 ```
+
+Then include it in the `rails_helper.rb`:
+
+```ruby
+config.include JavaScriptErrorCollector, type: :system
+config.after(:each, type: :system, js: true) { collect_errors }
```
macav, line 141: can we make it so, that it automatically collects errors for `js: true` as soon as we include the module? so we would have only `include` in the `rails_helper.rb`.
Also, wouldn't `JavascriptErrorReporter` be a bit better name? It doesn't collect it anywhere, it reports it on the console

Janis-Leuenberger:
Now it is as it was before, just in a module. It is included in the setup guide by default (no longer optional) and if someone doesn't like it they can just not do `js: true`. If someone dislikes the warnings or wants the test to fail, they can do it for their project.

ruby_on_rails/rspec.md
```diff
@@ -110,23 +110,10 @@ Commit and push your changes! :tada:
...
-        warn 'WARN: javascript warning'
-        warn error.message
-      end
-    end
-  end
-end
-```
+* Verify that `config.include JavaScriptErrorReporter, type: :system, js: true` is in your [`rails_helper.rb`](../templates/spec/rails_helper.rb)
+
+* Add `js: true` to any system test that should fail when JavaScript errors occur.
```
coorasse, line 119: that can be removed since it's implicit IMHO, and might lead to the bad behaviour of adding `js: tre` just to collect javascript errors, with the bad consequence of making the test 10x slower.

Janis-Leuenberger: IMHO (just learned that acronym) it doesn't make a difference. We can't stop anyone from doing that anyways. But I've still removed it. Hope we can finally close this?

sihu (approved):
Looks good to me now 👍

coorasse (approved):
Looks good to me

ruby_on_rails/rspec.md
schmijos: I don't like if people write "IMHO". It feels a bit like "with all due respect" to me.

PR #252, opened 2021-01-25
Point to .semaphore folder

phi-bre:
Small improvement to get directly to all semaphore files, since the text talks about it in the context the 'template', not just the entry file


phi-bre:
Ah, now I know why we did this. IMO we should remove [configure_cd](https://github.com/renuo/applications-setup-guide/blob/master/ruby_on_rails/configure_cd.md) and directly add the few relevant lines here.

(Eitherway, `Mobile apps are deployed via Bitrise.` needs to be removed)

schmijos (approved):

PR #254, opened 2021-02-02
Change SparkPost instructions to use subaccounts

macav:
After investigating the email providers situation, I think the subaccounts on Sparkpost are exactly what we need and we complicated our situation by setting up a completely new account for new projects.

image

Currently we have multiple projects that needed an upgrade to $20 plan, although they don't really utilise the plan. There's also quite a mess because of this, because some projects don't only have their own account, but are also set up under `sparkpost+renuoapp@renuo.ch` account.

By having them under one main account, we can have upgraded plan with (hopefully) better support on this account and covers all the subaccounts. Currently, there's 5 projects upgraded to a paid plan, but `renuoapp` is not one of them, so our customer support leverage is weaker.

Historically, I think the main reason to split the accounts was that there was 100 000 emails in the free plan, which is not the case anymore. 
 
Or can you recall if there was any other reason / problem with the subaccounts that I'm not aware of?

If this sounds good to you, I'd propose to migrate all the paying customers as subaccounts of the `renuoapp` account, which we upgrade.
Hopefully, this will not cause a lot of headaches with the sending domains, because they need to be detached from one account and verified in the new one.

### DIFF
--- a/sparkpost.md
+++ b/sparkpost.md
@@ -1,31 +1,24 @@
 # SparkPost
 
+Each app is using a separate subaccount under the main SparkPost account sparkpost+renuoapp@renuo.ch.
+
+There are two paths when setting up the project on SparkPost:
+
+* The domain is known: it should be set up and verified under the subaccount's sending domains.
+* The domain is not known: the emails will be sent with **\*@renuoapp.ch** as your mail sender
+
 :warning:
-**Never ever set up a *\*.renuoapp.ch* as a sending domain to a new created account. This will block the account on SparkPost.
-If you need this, choose the second option.** :warning:
+Never use the primary account for the project, so that the whole account doesn't get suspended / blocked in case of compliance issues!
 
-There are two options to set up SparkPost:
+There should only be one SparkPost subaccount per project. This means, that *master*, *develop* and *testing* use the same one.
 
-* With a domain for the project and a separate SparkPost account (preferred, if possible)
-* Without an exisiting domain ==> via renuoapp.ch SparkPost account
-
-**Note:** With the second option, you are only able to send mails with **\*@renuoapp.ch** as your mail sender.
-
-## First Option
-
-There should only be one SparkPost account per project. This means, that *master*, *develop* and *testing* use the same one.
-
-1. Go to <https://app.sparkpost.com/sign-up/> and create a new account
-1. Use the email `sparkpost+[project-name]@renuo.ch` as your username
-1. Use `renuo generate-password` to generate a secure password
-1. Copy the credentials into the credential store
-1. Fill in the domain you plan to use
-1. Confirm your email address (sent to sparkpost@renuo.ch)
-   > Hint: <https://groups.google.com/a/renuo.ch/forum/#!topic/mandrill>
-1. Write down the API-key in the credential store, because it's only shown once!
-1. You can check your key under: <https://app.sparkpost.com/account/credentials>
-1. Validate your sending domain [here](https://app.sparkpost.com/account/sending-domains), or add it if not already done (Set up SPF & DKIM with TXT DNS records)
+1. Go to <https://app.sparkpost.com/auth> and log in with the credentials for
+   sparkpost+renuoapp@renuo.ch found in the credential store
+1. Create [one new subaccount](https://app.sparkpost.com/account/subaccounts) and name it like your project
+1. Create [a new API-Key for your subaccount](https://app.sparkpost.com/account/api-keys/create) and assign it to the new subaccount, with the following permissions: *Send via SMTP, Sending Domains: Read/Write*
+1. Write down the API-key in the credential store, because it's only showed once!
 1. Credentials for SMTP setup on your app can be found [here](https://app.sparkpost.com/account/smtp), password is your generated API-key
+1. (if domain is known) Add your sending domain [here](https://app.sparkpost.com/domains/create?type=sending). Assign it to the subaccount. Set up SPF & DKIM with TXT DNS records
 1. Set up your ENV-variables and test if the mails are working. Manual test emails can be send via the following command in the rails console (production environment): `ActionMailer::Base.mail(to: 'yourname@renuo.ch', subject: 'Testmail', body: 'Mail content').deliver_now!`
 
 For DNS setup also see [Go Live](go_live.md)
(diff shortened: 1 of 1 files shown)


schmijos (approved):
Thank you very much!

I remember that we wanted to save money as the only reason for this complex setup. First we moved away from Mailchimp because it got more expensive and then Sparkpost changed its pricing as well. Bad choices if you look back.

macav:
In the meantime, I also opened a support ticket on SparkPost to confirm that in case of compliance issues, only the affected subaccount would be suspended and it wouldn't affect other customers.

coorasse:
What I recall was:
* sparkpost deactivated accounts/subaccount without notice. We got annoyed. We started migrating
* sparkpost didn't offer an easy heroku integration and therefore there were manual things to setup, and this was annoying
* we started testing sendgrid and we were very happy since it's a one-click solution on heroku

I personally would prefer a one-click (one heroku cli command) solution and a configuration per-heroku-project: this would give us more flexibility to select the most appropriate provider/plan/solution for each project.

I am a bit scared that:
* we are making a case for all our projects out of mtextur.
* we would be annoyed again by sparkpost.

macav:
I got a confirmation from SparkPost that the subaccounts are isolated and if one is sending abusive content, it will not affect the others. The only thing we need to be careful about is if the primary account is sending abusive content, this would block all subaccounts.

macav:
@coorasse but the deactivation without notice happened on SendGrid too. And these requirements pretty much are in contradiction to having one account with "tenant" system (subaccounts). 

For me, the one-click solution is not that important, because it takes anyway 3 minutes to generate the API key and set ENV variables and for production domain, you need additional steps to verify the sending domain anyway.

sparkpost.md
```diff
@@ -1,31 +1,24 @@
 # SparkPost
 
-:warning:
-**Never ever set up a *\*.renuoapp.ch* as a sending domain to a new created account. This will block the account on SparkPost.
-If you need this, choose the second option.** :warning:
-
-There are two options to set up SparkPost:
+Each app is using a separate subaccount under the main SparkPost account sparkpost+renuoapp@renuo.ch.
```
sihu, line 3:
```suggestion
Each app is using a separate subaccount under the main SparkPost account sparkpost@renuo.ch.
```
Could we simply go for this address?

sparkpost.md
```diff
@@ -1,31 +1,24 @@
...
 
-* With a domain for the project and a separate SparkPost account (preferred, if possible)
-* Without an exisiting domain ==> via renuoapp.ch SparkPost account
+There are two paths when setting up the project on SparkPost:
 
-**Note:** With the second option, you are only able to send mails with **\*@renuoapp.ch** as your mail sender.
+* The domain is known: it should be set up and verified under the subaccount's sending domains.
+* The domain is not known: the emails will be sent with **\*@renuoapp.ch** as your mail sender
```
sihu, line 8: i guess that was exactly the problem, that we sent emails on test and develop and therefore the account was blocked. this would affect the main account right?

sparkpost.md
```diff
@@ -1,31 +1,24 @@
...
 
-## First Option
+:warning:
+Never use the primary account for the project, so that the whole account doesn't get suspended / blocked in case of compliance issues!
 
-There should only be one SparkPost account per project. This means, that *master*, *develop* and *testing* use the same one.
+There should only be one SparkPost subaccount per project. This means, that *master*, *develop* and *testing* use the same one.
```
schmijos, line 13: This means, that we could break _master_ mail-sending by sending "abusive" emails from _develop_ and _testing_, right?

What about having one subaccount for all develop and one subaccount for all testing deployments? Something like sparkpost+develop@renuo.ch and sparkpost+testing@renuo.ch?

sparkpost.md
macav: true, if @renuoapp.ch is sending fishy content, this could affect others.
The only thing that comes to my mind would be to keep sparkpost+renuoapp@renuo.ch for develop / testing and have a separate, sparkpost@renuo.ch, for production

sparkpost.md
macav: I wrote somthing similar just now :) https://github.com/renuo/applications-setup-guide/pull/254#discussion_r569210519

coorasse: I'd propose to use mailtrap for testing and develop, but this is not always possible: it also depends on the customer.

macav: That's also fine for me, I can add it as an alternative. I think it's not necessary to split develop and testing sparkpost accounts, so I'd stick to either sparkpost+master@renuo.ch and sparkpost+develop@renuo.ch (which covers also testing), or keep sparkpost+renuoapp@renuo.ch as it is for develop / testing and set up sparkpost@renuo.ch as a main account for production

coorasse:
Yes, I don't find having everything under one account important if is an heroku extension.

I still am not sold on the "one account fits them all" solution. :)

We had problems in the past with Sparkpost, why won't we have problems in the future? 

IP Addresses: will all subaccounts use the same? Can it happen that if one client spams, they all end up in the spam folder? https://www.sparkpost.com/docs/deliverability/dedicated-ip-pools/

Sorry for being a bit picky here but it's because I care :)

macav:
@coorasse the problem with Heroku extension is, also with for example SendGrid, that they started enforcing 2FA. So this is a huge effort to maintain everything, if it's split across a million accounts. I'm receiving then 20 emails from SendGrid or Sparkpost, not knowing which app is actually affected, saying we have to enable 2FA (for example.. this was the case 1 month ago)

IP Adresses: Since also the compliance is isolated, I'd say that this should be the same also for the spam. But if not, then subaccounts don't have an affect on that. If that's an a concern, a dedicated IP pool can be always created and assigned to a subaccount: https://www.sparkpost.com/docs/deliverability/dedicated-ip-pools/#using-your-dedicated-ips-with-subaccounts

macav:
@coorasse we had some problems, yes, but we had them also with SendGrid and it's not guaranteed that we don't have them with other providers. 

But SparkPost is already a default in the setup guide and we still have it for majority of the apps, so my proposal doesn't change anything with that.

sparkpost.md
coorasse: I'd be more happy to keep a consistent naming. Yes, is not necessary to split testing and develop, but if it makes it easier to reason about it, then let's do it.

coorasse (approved):

sparkpost.md
macav: Ok, so to be clear:

* I'd change the address on sparkpost+renuoapp@renuo.ch to sparkpost+develop@renuo.ch
* I'll create sparkpost+testing@renuo.ch
* I'll create sparkpost+master@renuo.ch, where we will migrate and merge the separate accounts for clients as subaccounts.

sihu (approved):

sparkpost.md
schmijos: Perfect!

macav:
I created follow up tasks to carry out the migration and cleanup of the setup, as well as describe setup of Mailtrap in ASG as an alternative for develop & testing emails. Will merge this now.

sparkpost.md
coorasse: Good for me as well 👍

PR #259, opened 2021-02-26
add migration command to deploy script

pandermatt:


macav (approved):
aren't these templates also in the `renuo-cli`?

pandermatt:
I just checked, but they are not in the cli

coorasse:
@pandermatt https://github.com/renuo/renuo-cli/blob/master/lib/renuo/cli/app/templates/semaphore/semaphore-deploy.yml.erb

pandermatt:
Ahh it's a erb file, that's why I didn't found it. Thx

coorasse:
There's a search function, you can use for that 👍 
https://github.com/renuo/renuo-cli/search?q=git+push+heroku

PR #262, opened 2021-03-10
Raise on unpermitted params

schmijos:
Accidentially unpermitted params are a source of unnecessary errors and long bug-hunts. That's why we should raise whenever we receive information we don't wan't to receive in `development` and in `test`.


coorasse (approved):
This will not allow "partial updates" but can always be changed back if we need them

coorasse:
As a reference: https://github.com/rails/rails/pull/39310

PR #265, opened 2021-03-25
Update Sentry instructions

macav:


templates/config/initializers/sentry.rb
```diff
@@ -1,7 +1,10 @@
 # frozen_string_literal: true
 
-if defined?(Raven)
-  Raven.configure do |config|
-    config.processors -= [Raven::Processor::PostData]
```
macav, line 5: This is no longer supported, nor needed: https://docs.sentry.io/platforms/ruby/migration/#removed-processors

templates/config/initializers/sentry.rb
```diff
@@ -1,7 +1,10 @@
...
+if defined? Sentry
+  Sentry.init do |config|
+    filter = ActiveSupport::ParameterFilter.new(Rails.application.config.filter_parameters)
+    config.before_send = lambda do |event, _hint|
+      filter.filter(event.to_hash)
+    end
```
macav, line 8: I propose to add the parameter filtering as a default (https://docs.sentry.io/platforms/ruby/guides/rails/configuration/options/#optional-settings)

The `filter_parameter` config is defined by default in new projects in `config/initializers/filter_parameter_logging.rb`

schmijos (approved):
What about the environments?

macav:
@schmijos the environments were already documented before (using ENV variable `SENTRY_ENVIRONMENT`). Nothing changes there: https://github.com/renuo/applications-setup-guide/blob/master/ruby_on_rails/sentry.md#backend-rails

PR #266, opened 2021-04-07
Exclude routes from block length cop

lukasbischof:


templates/.rubocop.yml
```diff
@@ -40,6 +40,7 @@ Metrics/BlockLength:
   Exclude:
   - 'spec/**/*.rb'
   - 'config/environments/**/*.rb'
+  - 'config/routes.rb'
```
lukasbischof, line 43: Since routes usually grow anyways, ignoring it from the beginning probably makes sense

macav (approved):

coorasse (approved):
Fine for me. True that Rails 6.1 will make splitting them very easily, but since it feels more like a configuration file to me, I prefer to have all routes in one place rather than splitting them and make it harder to read/find them.

coorasse (approved):
Fine for me. True that Rails 6.1 will make splitting them very easily, but since it feels more like a configuration file to me, I prefer to have all routes in one place rather than splitting them and make it harder to read/find them.

schmijos (approved):

PR #267, opened 2021-04-22
Debug-Mode for capybara

sihu:
i really like this config and i would like it to be a default. Maybe there is another way how you do it?
What i like about it, is that you can run `DEBUG=true bin/check-system-specs` and it runs it in the browser (not headless)

### DIFF
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -92,7 +92,6 @@ You should know exactly why you are adding each one of them and why is necessary
   test:
     # SELENIUM_DRIVER: 'chrome'
     SELENIUM_DRIVER: 'headless_chrome'
-
   ```
  
   Please check the [rails_helper template](../templates/spec/rails_helper.rb).
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -93,7 +93,7 @@ You should know exactly why you are adding each one of them and why is necessary
     # SELENIUM_DRIVER: 'chrome'
     SELENIUM_DRIVER: 'headless_chrome'
   ```
- 
+
   Please check the [rails_helper template](../templates/spec/rails_helper.rb).
 
 * Add the line `bundle exec rspec` to `bin/check`
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -99,7 +99,9 @@ You should know exactly why you are adding each one of them and why is necessary
 
 * Add the line `bundle exec rspec` to `bin/check`
 
-> **Note**: If you want to debug a spec, you can simply run `rspec ./spec/my_spec.rb SELENIUM_DRIVER=selenium_chrome` to not run it headless or for a very verbose output use the `rspec ./spec/my_spec.rb DEBUG=true` 
+> **Note**: If you want to debug a spec, you can simply uncomment the line `SELENIUM_DRIVER` in the application.yml to not run it headless:
+  
+![CleanShot 2021-06-25 at 16 54 22](https://user-images.githubusercontent.com/1319150/123443347-1bbcae80-d5d6-11eb-8ba5-0d2c9ae4a37c.gif)
 
 ## :white_check_mark: Our first (green) test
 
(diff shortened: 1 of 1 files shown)


ruby_on_rails/rspec.md
```diff
@@ -78,8 +78,9 @@ You should know exactly why you are adding each one of them and why is necessary
       driven_by :rack_test
     end
 
+    SELENIUM_DRIVER = ENV['DEBUG'].present? && ENV['DEBUG'].eql?('true') ? :selenium_chrome : :selenium_chrome_headless
     config.before(:each, type: :system, js: true) do
-      driven_by :selenium_chrome_headless
+      driven_by SELENIUM_DRIVER
```
coorasse, line 83: I specify the driver directly in the environment variable. default to `:selenium_chrome_headless`

```
  driven_by ENV['SELENIUM_DRIVER'].to_sym
end
```

I find the ENV variable `DEBUG` too generic. With a different name I'd be ok, but I'd prefer this solution because it allows you to also run it on firefox for example if you need to. See also https://github.com/renuo/so_many_devices

sihu: @coorasse the advantage about `DEBUG` is, that selenium also knows this variable, so also its' output is verbose

ruby_on_rails/rspec.md
```diff
@@ -78,8 +78,9 @@ You should know exactly why you are adding each one of them and why is necessary
       driven_by :rack_test
     end
 
+    SELENIUM_DRIVER = ENV['DEBUG'].present? && ENV['DEBUG'].eql?('true') ? :selenium_chrome : :selenium_chrome_headless
```
macav, line 81: how about a combination of the two suggestions?

```suggestion
    SELENIUM_DRIVER = ENV['DEBUG'].present? && ENV['DEBUG'].eql?('true') ? :selenium_chrome : (ENV['SELENIUM_DRIVER']&.to_sym || :selenium_chrome_headless)
```

macav (approved):

ruby_on_rails/rspec.md
coorasse: it seems like another good reason to keep them separate, no?

coorasse: I never used the verbose output of selenium. How does it look like?

schmijos (approved):
I like `DEBUG`. It's a convention across a lot of software.

ruby_on_rails/rspec.md
sihu: it looks very verbose ;-) i kind of changed my opinion. It's rather unhandy since it outputs every single request. @schmijos then i would go for @coorasse 's solution

schmijos: It doesn't matter so much for me. I normally use `driven_by :selenium_chrome_headless` directly in a before-block because normally I don't want to run **all** the tests in the browser but only selected ones.

So yes, go on.

sihu: @coorasse if you are fine with this solution, i would merge

schmijos:
I think I changed my mind. I could use `DEBUG` just some days ago to show Tanny from KHW how we do stuff at Renuo (like browser tests). I always forget the driver names and I'd like very much if we would introduce a convenience env variable named like `RUN_IN_BROWSER=true` or similar. Something I can remember.

coorasse:
That's how I do it. I find it so easy to switch, that I have some doubts that we can make it any better:
CleanShot 2021-06-25 at 16 54 22

schmijos:
> That's how I do it. I find it so easy to switch, that I have some doubts that we can make it any better:
> CleanShot 2021-06-25 at 16 54 22

Ok. I think this good as well. But then let's make this a Renuo convention.

coorasse:
@sihu can you move this PR forward?

schmijos:
You can go a step further than @coorasse and adjust the config as well (since we use Selenium per default anyways):

```rb
# config/application.example.yml
test:
  # SELENIUM_DRIVER: 'chrome'
  SELENIUM_DRIVER: 'headless_chrome'

# spec/rails_helper.rb
    driven_by :selenium, using: ENV['SELENIUM_DRIVER'].to_sym, screen_size: [1280, 800]
  end
```

coorasse:
I'd not specify the screen size, still. I normally use https://github.com/renuo/so_many_devices to test on different sizes.

coorasse (approved):

PR #268, opened 2021-04-27
Update rubocop.yml

coorasse:
Since it counts as one for the method it should count as one also for the class. I don't want to be bother by the "class-level" check if the "method-level" one does not complain


macav (approved):

sihu (approved):
👍

pandermatt (approved):

PR #269, opened 2021-04-28
add rubocop to db folder

pandermatt:
I think its important to also have some linters in the migrations. We introduced linters to migration on https://github.com/renuo/loyalty and Edge and I liked it. Since it can still detect some in-reversible migrations or code smells

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -18,8 +18,8 @@ Style/Documentation:
 
 Style/MixinUsage:
   Exclude:
-  - 'bin/setup'
-  - 'bin/update'
+    - 'bin/setup'
+    - 'bin/update'
 
 Style/NonNilCheck:
   IncludeSemanticChanges: true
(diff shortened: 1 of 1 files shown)


templates/.rubocop.yml
```diff
@@ -36,11 +36,28 @@ Style/HashTransformValues:
 Style/HashAsLastArrayItem:
   Enabled: false
 
+Metrics/MethodLength:
+  CountAsOne:
+    - array
+    - hash
+    - heredoc
+  Exclude:
+    - 'db/**/*'
```
pandermatt, line 45: as suggested in https://github.com/renuo/applications-setup-guide/pull/268
I would also like to add CountAsOne to MethodLength and not only to ClassLength

templates/.rubocop.yml
```diff
@@ -18,8 +18,8 @@ Style/Documentation:
 
 Style/MixinUsage:
   Exclude:
-  - 'bin/setup'
-  - 'bin/update'
+    - 'bin/setup'
+    - 'bin/update'
```
pandermatt, line 22: I also changed the spacing to be consistent with the rest of the file

coorasse (approved):
agree :)

Liberatys (approved):
👍🏻 Noice

schmijos (approved):

PR #270, opened 2021-06-10
Disable MultipleMemoizedHelpers Cop

sihu:
[rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/MultipleMemoizedHelpers
](https://rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/MultipleMemoizedHelpers
)
The cop would check if example groups contain too many `let` and `subject` calls. As discussed with @coorasse this does not lead to better design in our opinion. People would change the amount of allowed let-statements or trick and use instance variables in `before`, if they need more. so i’d rather like to disable it.

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -89,6 +89,9 @@ RSpec/InstanceVariable:
 RSpec/NestedGroups:
   Max: 5
 
+RSpec/MultipleMemoizedHelpers:
+  Enabled: false
+
 AllCops:
   NewCops: enable
   Exclude:


macav (approved):

coorasse:
As a reference:
* https://github.com/rubocop/rubocop-rspec/issues/862
* https://thoughtbot.com/blog/lets-not
* https://github.com/rubocop/rspec-style-guide/issues/94

coorasse:
There seems to be a lot of discussions behind this topic. I am sure it was a good idea to disable it on one11 and possible on other projects where we have already multiple warnings due to this cop.
I proposed it to Simon to open a PR for the ASG but I didn't read the discussions entirely, so I don't have the feeling to be enough prepared on this topic. 
I'd just do it because I am used to use many lets and I don't want to change the way I write tests, but maybe there are interesting elements in the discussions that lead to this cop?

I'll take the time next weekend to read about the topic an give my opinion in a more "informed" way.
In the meantime, I have no problems in disabling it for existing projects because it might lead to too many hours of refactorings needed.

schmijos (approved):
GTM!

_Ceterum censeo standardrb esse introducam_

coorasse (approved):

PR #271, opened 2021-06-26
Add pry-doc as suggested

schmijos:
Makes Ruby docs accessible in pry.

Got the idea from here: https://www.shakacode.com/blog/railsconf-2021-implicit-to-explicit-decoding-rubys-magical-syntax/

### DIFF
--- a/suggested_libraries.md
+++ b/suggested_libraries.md
@@ -21,6 +21,7 @@ end
 group :development, :test do
   gem 'bullet'
   gem 'pry-byebug'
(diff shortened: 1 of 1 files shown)


coorasse (approved):
I barely use pry since I am on intellij, but I think this might be helpful for those who are not using intellij.

coorasse:
Actually is not true. I don't use it to debug, but I use it in the console

macav (approved):

PR #273, opened 2021-06-30
Fix sidekiq docs for async adapter and reformat

schmijos:
Using the [async adapter](https://api.rubyonrails.org/classes/ActiveJob/QueueAdapters/AsyncAdapter.html) in development seems to be closer to reality than `inline`.

### DIFF
--- a/ruby_on_rails/sidekiq.md
+++ b/ruby_on_rails/sidekiq.md
@@ -6,116 +6,123 @@ We use [sidekiq](https://github.com/mperham/sidekiq) because it works well on He
 
 * Add the following gem `gem 'sidekiq'` to the `production` block and install it
 
-```sh
-bundle install
-```
+  ```sh
+  bundle install
+  ```
 
-* update the `Procfile` and set the `worker` line, so it looks like this:
+* Update the `Procfile` and set the `worker` line, so it looks like this:
 
-```rb
-web: bundle exec puma -C config/puma.rb
-worker: bundle exec sidekiq -C config/sidekiq.yml
-```
+  ```rb
+  web: bundle exec puma -C config/puma.rb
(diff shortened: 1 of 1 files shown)


ruby_on_rails/sidekiq.md
```diff
@@ -6,116 +6,123 @@ We use [sidekiq](https://github.com/mperham/sidekiq) because it works well on He
...
+  ```
 
-* configure the active job adapter in `config/environments/development.rb`
+* Configure the active job adapter in `config/environments/development.rb`
 
-```rb
-config.active_job.queue_adapter = :inline
-```
+  ```rb
+  config.active_job.queue_adapter = :async
```
schmijos, line 29: This is the logical change.

macav (commented):
~~I think it was `inline` on purpose, as you don't need to be `sidekiq` running to see the async stuff happen in development (for example if the app only sends e-mails). I think that works fine until you have something more complex, such as `sidekiq-cron`.~~

Sorry, I just opened the documentation for that. From that, it seems to be quite close to `inline`, so fine for me.

schmijos:
> ~I think it was `inline` on purpose, as you don't need to be `sidekiq` running to see the async stuff happen in development (for example if the app only sends e-mails). I think that works fine until you have something more complex, such as `sidekiq-cron`.~
> 
> Sorry, I just opened the documentation for that. From that, it seems to be quite close to `inline`, so fine for me.

It uses `concurrent-ruby` to execute the jobs in parallel. `inline` executes them sequentially.

coorasse (approved):
I don't recall why we decided for `inline` and I don't see why not using `async`

sihu (approved):

PR #276, opened 2021-09-14
Staging Environment Protection

lukasbischof:
This should even work with XHR requests, since the browser sends the `Authorization` header during XHR calls, even without manual configuration. However, it could cause problems, when there's a mobile app which uses the applications api unless its served from an api controller that directly inherits from the rails api controller

### DIFF
--- a/ruby_on_rails/README.md
+++ b/ruby_on_rails/README.md
@@ -32,7 +32,7 @@ Most are not optional, but the rest of the team can start working even if those 
 3. [robots.txt](robots_txt.md)
 4. [Percy](configure_percy.md) (optional)
 5. [Papertrail](papertrail.md)
-6. [Protect staging environments](environment_protection.md)
+6. [Protect develop & testing environment](environment_protection.md)
 
 Some services should be configured accordingly to the packages bought by the customer.
 Once the new application is created, please add the project to the
--- a/ruby_on_rails/README.md
+++ b/ruby_on_rails/README.md
@@ -32,6 +32,7 @@ Most are not optional, but the rest of the team can start working even if those 
 3. [robots.txt](robots_txt.md)
 4. [Percy](configure_percy.md) (optional)
 5. [Papertrail](papertrail.md)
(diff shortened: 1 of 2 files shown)


ruby_on_rails/README.md
```diff
@@ -32,6 +32,7 @@ Most are not optional, but the rest of the team can start working even if those
 3. [robots.txt](robots_txt.md)
 4. [Percy](configure_percy.md) (optional)
 5. [Papertrail](papertrail.md)
+6. [Protect staging environments](environment_protection.md)
```
coorasse, line 35: "staging" is not really a word we commonly use. What about `Protect develop and testing environments`

ruby_on_rails/environment_protection.md
```diff
@@ -0,0 +1,26 @@
+# Staging Environment Protection
+
+To protect Renuo from a lawsuit for using copyrighted material, staging applications should be protected using HTTP Basic auth to prevent unwanted traffic.
```
coorasse, line 3: I'd not write this so explicitly here. I'd say that the main reason is that by doing so the develop and testing environments cannot be accessed publicly but only from us and our customers

ruby_on_rails/environment_protection.md
```diff
@@ -0,0 +1,26 @@
...
+
+  # ...
+end
+```
+
+Add `# BASIC_AUTH: 'admin:some-memorable-password'` to `application.example.yml` and run the following commands:
+
+```sh
+heroku config:set BASIC_AUTH='admin:[some-memorable-password]' --app [your-app]-develop
+heroku config:set BASIC_AUTH='admin:[some-memorable-password]' --app [your-app]-testing
```
coorasse, line 23: I'd specify to use two different ones and also to put them in the password manager.

coorasse (approved):

schmijos (approved):

schmijos:
I'm a bit afraid that this will complicate our and our customers' lifes. I hope we do this for security purposes and not for dyno hours…

PR #280, opened 2021-10-07
Update sentry configuration when customer does not pay

coorasse:
In order to not receive events (and therefore consume quota) for non-paying customers but still keep sentry configured, we can disable the key


sihu:
I see it the same as @schmijos. i think it would make us blind to not track those clients. I would appreciate if we could find a way to rate-limit per project. And if that is worth it, is a bit depending on the cost/value ratio.

PR #289, opened 2022-01-06
Add git flag

Janis-Leuenberger:
markdownlint would try to lint files inside `node_modules`. The `-g` flag allows just the linting of files known to git.

Screenshot 2022-01-06 at 14 24 13


coorasse (approved):
why do we even use mdl?

Janis-Leuenberger:
Mostly just the readme. I like having a markdown linter but not a must have by any means.

PR #293, opened 2022-01-11
Move factory_bot to proper group

Janis-Leuenberger:


schmijos (approved):

ruby_on_rails/rspec.md
```diff
@@ -6,12 +6,12 @@ Add the following gems to your Gemfile:
 
 ```ruby
 group :development, :test do
+  gem 'factory_bot_rails'
```
sihu, line 9: @Janis-Leuenberger, @schmijos the reason we had this in `:development, and :test` was very pragmatic: during development you can generate data very easily within the `rails console`. i don't think that is wrong. i would agree, that it shouldn't be in production (and therefore also not used in seeds).

coorasse: I use it also in the seeds :D

sihu: hmm...maybe i mixed something up. because there is also the `gem 'factory_bot'`. therefore @coorasse the point i brought up was not valid and i am fine with the changes.

Janis-Leuenberger: @sihu aren't you thinking of the faker discussion?

sihu: @Janis-Leuenberger no i ment factorybot. But it also uses faker sometimes and therefore goes into the same topic.

schmijos: The point of @Janis-Leuenberger is not about using factory bot but about the Rails generators, right?

PR #294, opened 2022-01-13
Fix status checks

Janis-Leuenberger:
One has to configure it like this:

Screenshot 2022-01-13 at 11 39 39

### DIFF
--- a/configure_git_repository.md
+++ b/configure_git_repository.md
@@ -18,7 +18,7 @@ Please stick to it unless you have special needs.
   * Default branch: `develop`. Click *update*
   * Add these rules for the two branches `develop` and `master`:
     * Require pull request reviews before merging
-    * Require status checks to pass before merging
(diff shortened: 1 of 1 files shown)


configure_git_repository.md
```diff
@@ -18,7 +18,7 @@ Please stick to it unless you have special needs.
   * Default branch: `develop`. Click *update*
   * Add these rules for the two branches `develop` and `master`:
     * Require pull request reviews before merging
-    * Require status checks to pass before merging
+    * Require status checks to pass before merging (after you configured the CI add it to the required checks)
```
Janis-Leuenberger, line 21: Ik, it's not the best spot but better than nothing, isn't it?

coorasse: Yes, what about mentioning something also in the CI/CD setup?

coorasse (approved):

PR #297, opened 2022-02-28
Adjust setup to new Ruby and asdf

schmijos:


ruby_on_rails/app_initialisation.md
```diff
@@ -2,15 +2,11 @@
...
-Follow the instructions on screen to eventually install it.
+* Install the latest Ruby version with `asdf install ruby latest` (Check if it's supported by Heroku).
 
-* Switch the global Ruby version to the newest one with `rbenv global <ruby-version>` so that the latter commands
-run in the correct context. You don't need to, if you've a `.ruby-version` in your projects containment folder.
+* Switch your global Ruby to the fresh one: `asdf global ruby latest`
 
-* Run `gem install bundler` to install Bundler.
-You may have it already installed, but this command will make sure that you have the latest version.
+* Run `gem update --system` to update Ruby's default gems (e.g. `bundler`).
```
schmijos, line 9: `bundler` is a default gem sind Ruby 2.6.0

sihu (approved):
LGTM 👍

ruby_on_rails/app_initialisation.md
```diff
@@ -19,14 +15,21 @@ You can do this with `gem update rails`. Beware of beta versions.
 You may want to choose a different database from Postgres, but most of the time that will be your choice.
 If you do not need a DB you may rethink the fact that you may not need Rails at all :) Take a look at [Sinatra](http://www.sinatrarb.com/) or [Angular](https://angular.io/)
 
+* Switch your local Ruby to the fresh one: `asdf local ruby latest` to populate a `.tool-versions` file
+  and delete any `.ruby-version` or `.nvmrc` files generated by Rails.
```
schmijos, line 19: Rails setups should not generate `.nvmrc` files. I still leave the hint.

ruby_on_rails/app_initialisation.md
```diff
@@ -19,14 +15,21 @@ You can do this with `gem update rails`. Beware of beta versions.
...
+
 * Load the Ruby version automatically in the fresh project's `Gemfile` by adding this:
 
-  ```ruby File.read(File.join(__dir__, '.ruby-version'))```
+  ```rb
+  ruby File.read(File.join(__dir__, '.tool-versions'))[/ruby\s(.*)/, 1].strip
```
schmijos, line 24: I thought about completely dropping this and replacing it by

```rb
ruby '~> 3.1.0'
```

since patch level updates never were a problem since Ruby 1.9.3. But I then thought, that we always want to go through the CI on a version change. So I now read the more complicated `.tool-versions` file.

coorasse: If Heroku and Semaphore are ok, I am ok 👍

coorasse (commented):
I have to be honest: switching away from standards "ruby-version" and "nvmrc" to go towards an `asdf` specific format, still does not convince me fully. I didn't find in the protocol why we decided in favour of that: https://docs.google.com/document/d/1kqig4ITDiSciXu6mGtIg8b0bUsqQXjNl8De6gu5Y3Hg/edit#heading=h.22jutnjr8ybt

Maybe we could have it documented there?

ruby_on_rails/app_initialisation.md
lukasbischof: FYI, There'd also be a way to configure asdf to use the `.ruby-version` file:
https://github.com/asdf-vm/asdf-ruby#migrating-from-another-ruby-version-manager
Still using `.ruby-version` maybe is easier since Heroku etc. already know it (maybe they also look at .tool-versions, I haven't checked 🤷)

schmijos: @lukasbischof the assumption that Heroku knows about it is wrong. Heroku looks at the Gemfile.

schmijos:
> Question: why not using asdf and keep the ruby-version and nvmrc files? 🤔

Why would you not want to have a single file with all virtual env dependencies in it?

schmijos:
There are two arguments against using `.tool-versions` right now:

* SemaphoreCI setup would need adjustment.
* dependabot doesn't read it

rnestler:
@schmijos @coorasse As part of looking into the CI cache issue we discussed in wg-engineering I'd also suggest to keep `.ruby-version` and set `legacy_version_file = yes` in `~/.asdfrc`.

coorasse:
It's a bit annoying that they call it `legacy`. 😆

rnestler:
> It's a bit annoying that they call it legacy. laughing

Yes it makes you feel bad for using it :wink:

PR #299, opened 2022-03-21
Update node version specification

Janis-Leuenberger:
https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version


odthoma (approved):
Thanks 👍🏻 I'd maybe also add that you should only specify the major version. (e.g `16.x`) We should still keep the `.nvmrc` btw.

Also maybe that you'd have to add the nodejs buildpack on heroku in addition to the ruby one => in the correct order

(ah well i guess this is for the CI??)

Janis-Leuenberger:
1. I agree on just adding the major version.
2. Why do you want to keep the `.nvmrc`?
3. Heroku will figure out the correct buildpack most of the time. In the other cases you will know soon enough.
4. This is for heroku.

Do we even need node in production?

schmijos (changes requested):
Oi, Heroku is not to be mixed-up with Sempaphore:

> The version of Node.js that will be used can be set from a .nvmrc file, if such a file exists in your repository.

https://docs.semaphoreci.com/programming-languages/javascript-and-node-js/

odthoma:
Well, AFAIK we decided upon still sticking to `.nvmrc` for specifying the node version instead of `.tool-versions`. Thus we should still keep it. I still want my version manager to automatically switch my node environment. And yes, we only need node to compile our assets on production I guess.

Janis-Leuenberger:
So this part should be kept as is and add the changes to the application server setup instead?

Still we will have to specify the node version twice even if we go with the `nvmrc`. Opinions, solutions?

coorasse:
`.nvmrc` is still needed locally and semaphore. isn't there a way to have heroku reading it as well? 🤷

odthoma:
@coorasse I think we already talked about this once and didn't find a proper solution. But I mean how often do you really change the node version. Having to change it in 2 spots it not that bad.

We could maybe add a few lines to `bin/fastcheck` that would throw an error when the version in the `package.json` and in the `.nvmrc` are not lined up.

coorasse:
@Janis-Leuenberger please find an agreement. and try to finalise this :)

PR #174, opened 2019-04-03
Update Rails initialisation (shorten things)

schmijos:

PR #300, opened 2022-03-26
Update app initialization guide to support rails 7 setup

odthoma:


ruby_on_rails/README.md
```diff
@@ -57,7 +57,6 @@ useful services:
 1. [Amazon S3 and Cloudfront](aws.md)
 1. [Carrierwave Upload](carrierwave.md)
 1. awesome_print `gem 'awesome_print'`
-1. [bootstrap](bootstrap.md)
```
odthoma, line 60: Since you can now just use the `--css` flag or run `rails css:install:bootstrap` this is not needed anymore

ruby_on_rails/app_initialisation.md
```diff
@@ -72,7 +71,22 @@ They are always idempotent (runnable multiple times).
...
   Also change `system!('bundle install')` to `system!('bundle install --jobs=3 --retry=3')`
-  and uncomment the installation of yarn dependencies with `system('bin/yarn')`
+
+#### If you are using `importmap-rails`
+
+* Add a `bin/dev` file and make it executable. It will be used to start our project.
+
+  ```sh
+  echo "#\!/bin/sh\nset -e\n\nrails s" >> bin/dev && chmod +x bin/dev
+  ```
```
odthoma, line 81: To make things uniform between projects using importmaps or node, I'd change the name of the script to `bin/dev`

ruby_on_rails/app_initialisation.md
```diff
@@ -11,11 +11,16 @@
 * [Check if you are using the latest stable version of Rails](http://rubyonrails.org/) with `rails -v` and update it if you are not.
 You can do this with `gem update rails`. Beware of beta versions.
 
-* Start a new Rails project using `rails new [project-name] --database=postgresql --no-skip-test --skip-sprockets --webpack --skip_action_mailbox` where the `project-name` is exactly the one you chose before.
-You may want to choose a different database from Postgres, but most of the time that will be your choice.
+* Start a new Rails project using `rails new [project-name] --database=postgresql --no-skip-test --skip-action-text --skip-action-mailbox` where the `project-name` is exactly the one you chose before.
```
odthoma, line 14: Why is `--no-skip-test ` used here? I mean we use RSpec anyways.

ruby_on_rails/app_initialisation.md
```diff
@@ -11,11 +11,16 @@
 * [Check if you are using the latest stable version of Rails](http://rubyonrails.org/) with `rails -v` and update it if you are not.
 You can do this with `gem update rails`. Beware of beta versions.
 
-* Start a new Rails project using `rails new [project-name] --database=postgresql --no-skip-test --skip-sprockets --webpack --skip_action_mailbox` where the `project-name` is exactly the one you chose before.
-You may want to choose a different database from Postgres, but most of the time that will be your choice.
+* Start a new Rails project using `rails new [project-name] --database=postgresql --skip-test --skip-action-mailbox` where the `project-name` is exactly the one you chose before.
```
odthoma, line 14: y was `--no-skip-test` used here before? I mean we manually delete the generated test stuff anyways because of RSpec

ruby_on_rails/app_initialisation.md
```diff
@@ -72,7 +71,22 @@ They are always idempotent (runnable multiple times).
...
+  system! 'yarn install'
+  ```
+
+#### If you use `importmap-rails` and there is no `bin/dev` file yet
+
+* Create one and make it executable. It will be used to start our project.
+
+  ```sh
+  echo "#\!/bin/sh\nset -e\n\nrails s" >> bin/dev && chmod +x bin/dev
+  ```
```
odthoma, line 89: When using forman (i.e. when using `jsbundling-rails`) the default generated script for running your app is `bin/dev`. Therefore I'd also change it to `bin/dev` for projects using importmaps. That way it is consistent between new projects.

ruby_on_rails/README.md
coorasse: Two things:
* I am not very happy with Rails defaults. Using css:install:bootstrap also installs js and css bundling when is actually not necessary. See also https://dev.to/coorasse/rails-7-bootstrap-5-and-importmaps-without-nodejs-4g8
* That page contains also simpleform instructions

ruby_on_rails/app_initialisation.md
coorasse: I am good with it. @schmijos you did put some thoughts into that in the past. Any remark?

ruby_on_rails/app_initialisation.md
```diff
@@ -11,11 +11,16 @@
 * [Check if you are using the latest stable version of Rails](http://rubyonrails.org/) with `rails -v` and update it if you are not.
 You can do this with `gem update rails`. Beware of beta versions.
 
-* Start a new Rails project using `rails new [project-name] --database=postgresql --no-skip-test --skip-sprockets --webpack --skip_action_mailbox` where the `project-name` is exactly the one you chose before.
-You may want to choose a different database from Postgres, but most of the time that will be your choice.
+* Start a new Rails project using `rails new [project-name] --database=postgresql --skip-test --skip-action-mailbox` where the `project-name` is exactly the one you chose before.
+
+  * You can use the `--css` flag to preconfigure your new Rails app to use a CSS processor/framework. Available options are: `tailwind, bootstrap, bulma, postcss, sass`. This will automatically install the [cssbundling-rails](https://github.com/rails/cssbundling-rails) and [jsbundling-rails](https://github.com/rails/jsbundling-rails) gems. The app then uses the the [esbuild](https://github.com/evanw/esbuild) JS bundler and minifier by default.
```
coorasse, line 16: This is in the Rails guide. No need to repeat concepts here.

ruby_on_rails/app_initialisation.md
```diff
@@ -11,11 +11,16 @@
...
+
+  * Esbuild is generally a good choice, but you can also configure an alternative JS bundler using the `--javascript` flag. You have the choice between using one of the node-based bundlers: `esbuild, rollup, webpack` or you can use `importmap`, which is the default. Note that importmaps cannot be used in conjunction with the `--css` option from above.
```
coorasse, line 18: * This also seems part of the Rails guide. 
* These decisions are taken by the team anyway, this paragraph does not give guidance. I'd rather state at the very beginning that the team must decide before setting up the app if they want a js bundling and which one.
* Given that I'd even restrict these decision to always use esbuild
* Given that, I'd even restrict it more to say: "no js bundling until you really need it
* Why cannot importmaps be used with `--css`? 🤔

ruby_on_rails/app_initialisation.md
```diff
@@ -11,11 +11,16 @@
...
+
+  * You may want to choose a different database from Postgres, but most of the time that will be your choice.
```
coorasse, line 20: I'd add also here that a setup different from postgres. should be discussed and agreed with the team before setting up the app.

ruby_on_rails/app_initialisation.md
```diff
@@ -11,11 +11,16 @@
...
 If you do not need a DB you may rethink the fact that you may not need Rails at all :) Take a look at [Sinatra](http://www.sinatrarb.com/) or [Angular](https://angular.io/)
 
-* Check if you've got a `.ruby-version` file. If not, create one and enter the Ruby version.
+* Check if you've got a `.ruby-version` file and create one if not . If  there is already one in the project, be sure to remove the `ruby-` prefix from the version number inside the file, leaving only the number itself.
```
coorasse, line 23: Why is this necessary?

ruby_on_rails/app_initialisation.md
```diff
@@ -29,8 +34,8 @@ If you do not need a DB you may rethink the fact that you may not need Rails at
 
 * Run `bundle exec rails db:migrate` to generate an empty `schema.rb` file.
 
-* Then check your default Rails setup by running `rails s` and visiting `localhost:3000`.
-  You should be on Rails now, yay!
+* Check your default Rails setup by running either `bin/dev` or `rails s`, depending on whether you are using a node-based bundler with the foreman Procfile-runner, or if you are using importmaps and then visit `localhost:3000`.
```
coorasse, line 37: At this point of the story, `bin/rails s` will work anyway since there are no assets in the homepage, right?

ruby_on_rails/app_initialisation.md
```diff
@@ -39,12 +44,6 @@ If you do not need a DB you may rethink the fact that you may not need Rails at
 The following scripts are standardized tools for more convenience at Renuo.
 They are always idempotent (runnable multiple times).
 
-* Add a `bin/run` file. It will be used to start our project.
```
coorasse, line 42: why is this gone? 😢

ruby_on_rails/app_initialisation.md
coorasse: I'd keep consistency with all our previous project and go with a `bin/run` that either calls `bin/dev` or `bin/rails s`

ruby_on_rails/app_initialisation.md
```diff
@@ -137,18 +151,7 @@ create the initializer for figaro in `config/initializers/figaro.rb`:
 * Enable the default [Content Security Policies](https://github.com/renuo/applications-setup-guide/blob/master/ruby_on_rails/content_security_policy.md) in `config/initializers/content_security_policy.rb`.
   The report URI will be set later in the step of Sentry configuration.
 
-* If you're using Webpacker, let's clean up after asset precompilation
```
coorasse, line 140: Isn't this necessary somehow also when using esbuild, webpack or other js bundling systems?

coorasse (changes requested):

coorasse:
I asked for different reviews since this will affect all of us. Is really important so I'd like an opinion from almost everyone

ruby_on_rails/app_initialisation.md
odthoma: No because sprockets & propshaft provide this rake task by default. 

```ruby
rails assets:clean                       # Removes old files in config.assets.output_path
rails assets:clobber                     # Remove config.assets.output_path
```

ruby_on_rails/app_initialisation.md
odthoma: Yes that's a good idea. Let's keep it around.

ruby_on_rails/app_initialisation.md
odthoma: By default the version is now prefixed with `ruby-` (e.g` ruby-3.1.1`). 

That means that this won't work anymore in the Gemfile:
```ruby
ruby File.read(File.join(__dir__, '.ruby-version')).strip
```

ruby_on_rails/README.md
odthoma: 1. That's y I mentioned it in the [app_initialisation.md](https://github.com/renuo/applications-setup-guide/pull/300/files#diff-d3b0ddcebb15ddc57c8019594216282911df2799e742fdc977c3a7ecb624544d)
2. The simpleform instructions on that page aren't needed anymore because you can do `rails g simple_form:install --bootstrap.`

ruby_on_rails/app_initialisation.md
odthoma: Yes I just tried it, thanks

ruby_on_rails/app_initialisation.md
odthoma: - I still feel like this is a good place to mention that this is possible. Especially when the bootstrap guide is gone.
- I agree with giving more guidance and restricting it to esbuild. Also for consistency reasons I guess.
- When you use `--css` it will always default to esbuild. Even if you'd say `-c bootstrap -j importmap`

ruby_on_rails/app_initialisation.md
schmijos: I find it very weird that I added it here: https://github.com/renuo/applications-setup-guide/pull/174

I cannot remember why. Could it be that it comes with a Capybara and Selenium setup which we would have to do manually otherwise?

Generally: I'm fine with the suggested change.

schmijos: Yes, I guess, the reason is exactly what I thought. There are default configuration which we probably would like to keep:

CleanShot 2022-03-28 at 10 31 50

ruby_on_rails/app_initialisation.md
schmijos: `asdf` reads it correctly. I suggest, we add a helper class for such Renuo customization things. What do you think?
CleanShot 2022-03-28 at 10 34 09

ruby_on_rails/app_initialisation.md
schmijos: 😭

schmijos (commented):
I'm generally also on the side to with importmaps per default.

ruby_on_rails/app_initialisation.md
Liberatys: > You may want to choose a different database from Postgres, but most of the time that will be your choice.

Is this really needed? Postgres should be the default unless otherwise specified by the client? The only exception to this is probably using something like Firebase? Which is an entirely different discussion.

ruby_on_rails/app_initialisation.md
Liberatys: I'd like to keep it as asdf compatible as possible :D

ruby_on_rails/app_initialisation.md
Liberatys: TBH: I don't mind running bin/dev ^^ alias br = 'bin/run' => alias br = 'bin/dev'

ruby_on_rails/app_initialisation.md
lukasbischof: I think it's also convenient to keep `bin/run` since we have it also for non-rails projects. For example, I can run `bin/run` in the Renuo website project or a flutter app without even knowing the correct elixir/flutter command. 
Renaming it in all these projects seems bit weird just because the Rails standard changed :)

ruby_on_rails/app_initialisation.md
```diff
@@ -72,7 +71,22 @@ They are always idempotent (runnable multiple times).
...
 
   Also change `system!('bundle install')` to `system!('bundle install --jobs=3 --retry=3')`
-  and uncomment the installation of yarn dependencies with `system('bin/yarn')`
+
+#### If you are using a node-based JS bundler using `jsbundling-rails`
+
+* Add the following to `bin/setup` to allow intalling all dependencies in one go
+
+  ```ruby
+  system! 'yarn install'
```
lukasbischof, line 80: Shouldn't it actually be `bin/yarn install --check-files`? That's at least the recommended command in the error message of Rails when you're dependencies are out of date

ruby_on_rails/app_initialisation.md
rnestler: You even made a comment about it on the pull request: https://github.com/renuo/applications-setup-guide/pull/174/commits/359b3e2ce748c96ee60c456b623b4f7f06c17d49

> --no-skip-test should make clear that we want the test setup to be genereated even though we want to use RSpec later.

ruby_on_rails/app_initialisation.md
ignaciosy: +1

Janis-Leuenberger:
I'm on the side of import maps but more important to me is that setup guide works flawlessly. Last time I had issues with rails 7 js/css setup so please take the time to verify the guide.

Liberatys (approved):
LGTM

coorasse:
@odthoma please move forward with this PR.

Janis-Leuenberger:
@odthoma

ruby_on_rails/app_initialisation.md
rnestler: What I think is worse, is that we now have projects with `bin/run` and `bin/dev` which sometimes do different stuff (bin/run just does rails s and bin/dev starts foreman)

PR #302, opened 2022-04-19
feat: Migrate linting from semaphore to github actions

Liberatys:
Migrate our current md linting system from semaphore/ruby to GitHub actions with mdlint provider.

1. Free of charge. For public repos there is currently no cost for using Github actions
2. No more bin/
3. No ruby installation in our 'docs'
4. Still uses our configuration for mdlint


.github/workflows/linting.yml
```diff
@@ -0,0 +1,18 @@
...
+      - master
+jobs:
+  lint:
+    runs-on: ubuntu-20.04
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run mdl
+        uses: actionshub/markdownlint@main
```
Janis-Leuenberger, line 18: Is it a standard practice to point to a branch on github? Are there no releases or such?

Janis-Leuenberger (approved):
I like it

.github/workflows/linting.yml
Liberatys: That actually depends on the org ^^ actionshub is usually directly using their master branch as they ensure that is in a usable state.

Both referencing branches and versions are widespread.

.github/workflows/linting.yml
```diff
@@ -0,0 +1,18 @@
+name: Checks
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
rnestler, line 9: A bit late to the party, but why only on master and not main, development as well? For the `pull_request` I'd leave the branches filter off completely, since why would we want to limit it on pull_request to master?

.github/workflows/linting.yml
rnestler: >  actionshub is usually directly using their master branch as they ensure that is in a usable state.

But at the same time they seem to be using semver for their tags? So shouldn't we still pin on the major version to avoid breaking CI due to an action change?

.github/workflows/linting.yml
Liberatys: 👍🏻 although I'd love to use GitHub actions for our other projects I think we'll stay with semaphore for some time 😆

PR #303, opened 2022-04-25
Update gh-pages.yml

coorasse:


rnestler (changes requested):
Need to also update `.github/workflows/linting.yml` which still references `master`.

Liberatys (approved):

rnestler (approved):

PR #306, opened 2022-06-20
Mention to set 'Include administrators'

rnestler:
Since the staff team has administrator access we need to include them.


schmijos:
This is intentionally unconfigured. I'd rather have a record of force-merged PRs instead of people acting directly on `develop` to circumvent quality measures occasionally. There should not be any argument to circumvent PRs.

PR #309, opened 2022-09-26
Enforce ~consistent~ either hash syntax

schmijos:
`EnforcedShorthandSyntax: always` is the default setting in Rubocop even though Bozhidar himself [doesn't like the option](https://batsov.com/articles/2022/01/20/bad-ruby-hash-value-omission/). 

~I suggest to change the default option to `consistent` because I can see the advantage of the shorthand syntax when passing on options to deeper levels. But in any other context I found myself to be confused by it.~

~This change allows hashes to be in shorthand syntax only if the whole hash uses it consistently. See https://github.com/rubocop/rubocop/pull/10776.~

**Update:**
I changed my mind. Let's follow standardrb. Be consistent yourselves if you see fit.

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -27,6 +27,10 @@ Style/NonNilCheck:
 Style/HashEachMethods:
   Enabled: true
 
+Style/HashSyntax:
+  Enabled: true
+  EnforcedShorthandSyntax: consistent
+
 Style/HashTransformKeys:
   Enabled: true
 
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -29,7 +29,7 @@ Style/HashEachMethods:
 
 Style/HashSyntax:
   Enabled: true
-  EnforcedShorthandSyntax: consistent
+  EnforcedShorthandSyntax: either
 
 Style/HashTransformKeys:
   Enabled: true


rnestler (approved):

rnestler:
> though Bozhidar himself [doesn't like the option](https://batsov.com/articles/2022/01/20/bad-ruby-hash-value-omission/).

After reading this, I think I'd even vote to disallow it completely. But consistent style is probably enough.

coorasse (commented):
Can somebody explain to me the difference between `consistent` and `either`? Because I enabled the second one on vista.

odthoma:
@coorasse 

<details>
<summary>EnforcedShorthandSyntax: consistent</summary>

```
# bad
{foo: , bar: bar}

# good
{foo:, bar:}

# bad
{foo: , bar: baz}

# good
{foo: foo, bar: baz}
```

</details>

<details>
<summary>EnforcedShorthandSyntax: either</summary>

```
# good
{foo: foo, bar: bar}

# good
{foo:, bar:}
```

</details>

coorasse:
I tried to run it with `consistent` on vista currently and I get the following for example:

```
C: [Correctable] Style/HashSyntax: Omit the hash value.
        let(:section_instance) { section.new(card_request: card_request) }
```

I don't want to be forced to write `let(:section_instance) { section.new(card_request: card_request) }`.

At the same time it does not complain about `some_method(:hello, :b, card_request:, logger:)`.

coorasse (changes requested):
I tried to run it with consistent on vista currently and I get the following for example:

```
C: [Correctable] Style/HashSyntax: Omit the hash value.
        let(:section_instance) { section.new(card_request: card_request) }
```

I don't want to be forced to write `let(:section_instance) { section.new(card_request: card_request) }`.

At the same time it does not complain about `some_method(:hello, :b, card_request:, logger:)`.

As of now, I'd prefer `either` to `consistent`

coorasse:
I found also an old discussion we had. Maybe irrelevant now: https://renuo.slack.com/archives/C013FEGN4D7/p1650550992375259

coorasse:
either is also the default on standardrb: https://github.com/testdouble/standard/blob/main/config/base.yml#L1358

odthoma (approved):

coorasse (approved):

PR #312, opened 2022-10-03
Update CI setup instructions to use renuobot

firemind:


firemind:
> Why do we need to use renuobot and not our personal account?

2 reasons. First it will link and authenticate that semaphore project with your account so if you ever change the password or leave the company it needs to be fixed manually. The other issue is that it requires you to be an organisation owner in oder to set up the semaphore 2.0 integration.

rnestler (approved):

sihu (approved):

PR #314, opened 2022-10-12
Disable RSpec/NoExpectationExample

coorasse:
Too many false positives. We use custom expectations very often and the risk of having a test without an expectation is low: https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecnoexpectationexample

This cop simply goes too far.


lukasbischof (approved):
I think we even started excluding the cop (in Swissstaffing for example). I'm happy if we just exclude it in general

PR #316, opened 2022-10-24
Disable RSpec/SortMetadata check

rnestler:
The current order of the specified metadata makes more sense by putting the common options first and the special ones last. Sorting them alphabetically makes it less readable.

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -109,6 +109,9 @@ RSpec/MultipleMemoizedHelpers:
 RSpec/NoExpectationExample:  
   Enabled: false
 
+RSpec/SortMetadata:
+  Enabled: false
+
 AllCops:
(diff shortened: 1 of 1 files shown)


coorasse (approved):

sihu:
i will add your example for context here to follow the discussion: 

> `RSpec/SortMetadata`: Sort metadata alphabetically. `config.before(:each, type: :system, js: true, desktop: true) do` should be sorted. I don't like that one, since the order now is the common "default" options first and the special `desktop: true` option in the end which I think makes more sense than sorting alphabetically.

sihu (approved):

rnestler:
CI broke since we use the `main` branch for the markdown linting. See https://github.com/renuo/applications-setup-guide/pull/318

PR #318, opened 2022-10-24
Pin markdownlint version to 0.11.0 to fix CI

rnestler:
Just using the latest version of the main branch will break our CI workflow all the time.


rnestler:
I tried to just pin the version of markdownlint, but apparently it just installs the latest version of mdl. So we just install the old mld version which still works.

.github/workflows/linting.yml
```diff
@@ -14,7 +14,9 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
-      - name: Run mdl
-        uses: actionshub/markdownlint@main
+      - uses: ruby/setup-ruby@v1
         with:
-          filesToIgnoreRegex: "SUMMARY.md"
+          ruby-version: '3.1'
+
+      - run: gem install mdl --version 0.11.0
```
coorasse, line 21: Why is this gem not specified in the Gemfile?

rnestler: Because there is no Gemfile. I can add one if that would be cleaner.

schmijos: Project doesn't have a Gemfile nor a Ruby version specified.

schmijos: Yes, please add one. Otherwise the CI may lint differently than I do locally.

rnestler: Maybe it would indeed be clever to add a Gemfile and a `bin/check` script so one could also run mdl locally?

coorasse: Yes. Agree to both Gemfile and bin/check

rnestler: Done.

Gemfile
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+source 'https://rubygems.org'
+
+group :test do
+  gem 'mdl', '~> 0.11.0'
```
coorasse, line 6: I'd not lock the mdl version here. This is already locked in the Gemfile.lock.

rnestler: Hmm. I'd argue that this makes it more explicit that it will currently break if upgraded but on the other hand we probably want to upgrade at some point and fix the issues.

rnestler: Fixed.

coorasse (approved):

PR #319, opened 2022-11-03
Disable RSpec/AnyInstance

coorasse:
I am really annoyed by this cop. If I am using `allow_any_instance` is because I want to use it and doing it differently is just painful.

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -111,6 +111,9 @@ RSpec/NoExpectationExample:
 
 RSpec/SortMetadata:
   Enabled: false
+  
+RSpec/AnyInstance:
+  Enabled: false
 
 AllCops:
   NewCops: enable


firemind (approved):
I agree with this. Having to refactor code to expose instances just so they are testable like that will lead to unnecessary abstraction and a worse code-base overall. Obviously the feature should be used with some common sense but forbidding it all together is not the way to go.

Janis-Leuenberger (approved):

ignaciosy (commented):
I understand the pain, though also, I think without this rule I would never have investigated about other options, like instance doubles. Yes it can be annoying, but I'd be ok with placing the explicit disable-rule comment when i want to use it.

schmijos (approved):
Agreed. That's why have Ruby in the first place. Power in a few lines of code.

PR #322, opened 2022-12-08
Disable RSpec/VerifiedDoubles

coorasse:
https://relishapp.com/rspec/rspec-mocks/docs/verifying-doubles I agree that is important but I cannot always do it.

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -114,6 +114,9 @@ RSpec/SortMetadata:
   
 RSpec/AnyInstance:
   Enabled: false
+  
+RSpec/VerifiedDoubles:
+  Enabled: false  
 
 AllCops:
   NewCops: enable


schmijos (approved):
I guess we'll have a new learning week block: "Run Vanilla Rubocop on a project and investigate all offenses to learn about them."

lukasbischof:
Is it really wise to disable it for _all_ projects? If you say "not always" then this means mostly yes so why not simply disable it explicitly where it makes sense?

coorasse:
Because I am aware of how to use it. But I understand if this can be used as training.

rnestler:
Hmm I assumed since you ignored it globally on vista (or was it cobra) that it is generally not a valuable lint. But if it is generally valuable it favor to annotate it explicitly for the exceptional cases where it needs to be violated.

coorasse:
Not many more opinions during these five days, I'll go with it.

PR #325, opened 2023-01-04
Disable Rails/BulkChangeTable

coorasse:
Because "who cares?" (semicit.)

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -9,6 +9,9 @@ Rails/SkipsModelValidations:
 
 Rails/WhereExists:
   Enabled: false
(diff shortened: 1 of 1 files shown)


firemind (approved):
Guess this makes sense for mysql where schema changes are not transactional and you don't  want half the changes to be made if later ones fail.

schmijos:
> Because "who cares?" (semicit.)

"who cares" was intended for the Rubocop adjustment in a very old and irrelevant migration.

coorasse:
Ah! I actually thought about it for new migrations. I think old migrations should be removed anyway, so... :)

PR #326, opened 2023-01-04
Disable Bundler/OrderedGems

coorasse:
Because why do you care? If you want to check if a gem is configured you look at Gemfile.lock anyway


firemind (approved):
Agreed. Also the insanity that is this https://github.com/renuo/herold-planner/blob/develop/Gemfile#L54-L55

schmijos (approved):

PR #330, opened 2023-01-23
Disable buggy ActionControllerFlashBeforeRender cop

rnestler:
This cop seems to have quite a lot false positives pop up every now and then which are actually harmful since we won't render the flash in these cases.

See https://github.com/rubocop/rubocop-rails/issues/843 and https://github.com/renuo/vdrb-kas/pull/236/commits/7978dd7518f22fcef6768fe8fceed9f56377cddf

### DIFF
--- a/templates/.rubocop.yml
+++ b/templates/.rubocop.yml
@@ -68,6 +68,9 @@ Rails/I18nLocaleAssignment:
   Exclude:
     - 'spec/**/*'
 
+Rails/ActionControllerFlashBeforeRender:
+  Enabled: false
+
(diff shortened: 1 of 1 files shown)


coorasse (approved):
Agree. Please update also https://github.com/renuo/renuocop/blob/main/config/base.yml

rnestler:
> Agree. Please update also https://github.com/renuo/renuocop/blob/main/config/base.yml

This gives me a 404?

coorasse:
uops. now not anymore!

coorasse (approved):

PR #336, opened 2023-07-18
Add VCR example configuration

schmijos:


ruby_on_rails/vcr.md
```diff
@@ -1,8 +1,52 @@
...
+VCR.configure do |c|
+  c.hook_into :webmock
+  c.configure_rspec_metadata!
+  c.cassette_library_dir = 'spec/vcr'
+  c.default_cassette_options = {
+    decode_compressed_response: true,
+    allow_unused_http_interactions: false,
+  }
+  c.debug_logger = $stderr if ENV['DEBUG'] == 'true'
+  c.ignore_localhost = true
```
schmijos, line 36: I needed this to workaround the issue described here: https://stackoverflow.com/questions/74809209/webmock-stopping-selenium-from-deleting-session
`WebMock.disable_net_connect!(allow_localhost: true)` was not enough.

ruby_on_rails/vcr.md
```diff
@@ -1,8 +1,56 @@
...
+  c.debug_logger = $stderr if ENV['DEBUG'] == 'true'
+
+  c.hook_into :webmock
+  c.ignore_localhost = true
+
+  c.cassette_library_dir = 'spec/vcr'
+  c.default_cassette_options = {
+    decode_compressed_response: true,
+    allow_unused_http_interactions: false,
+    record: ENV['VCR'] ? ENV['VCR'].to_sym : :once, # re-record with VCR=all
```
firemind, line 38:
```suggestion
    record: ENV.fetch('VCR_RECORD', :once).to_sym, # re-record with VCR_RECORD=all
```

I'd prefer to be a bit more explicit in the variable name since just `VCR=all` can easily be misinterpreted as "have all VCRs active".

firemind (approved):
Just a rename suggestion. Otherwise this looks good to me.

ruby_on_rails/vcr.md
```diff
@@ -1,8 +1,56 @@
...
+  config.around do |example|
+    if example.metadata[:vcr]
+      example.run
+    else
+      VCR.turned_off { example.run }
+    end
+  end
+end
+
+# WebMock catches everything per default, we allow localhost for Capybara/Selenium
```
sihu, line 24: and maybe also to not mock away internal requests?

ruby_on_rails/vcr.md
```diff
@@ -1,8 +1,56 @@
...
+  # Filter out sensitive data from the cassettes
+  env_keys = YAML.load_file('config/application.example.yml').filter do |k, v|
+    (!k.in? %w[test production development]) && (!v.in? %w[false true])
+  end.keys
+  env_keys.each { |key| c.filter_sensitive_data("<#{key}>") { ENV.fetch(key, nil) } }
+end
+```
+
+Some considerations:
+* Do you really want/need VCR? A fake may be better: https://thoughtbot.com/blog/how-to-stub-external-services-in-tests#create-a-fake-hello-sinatra
```
sihu, line 51:
```suggestion

* Do you really want/need VCR? A fake may be better: https://thoughtbot.com/blog/how-to-stub-external-services-in-tests#create-a-fake-hello-sinatra
```

ruby_on_rails/vcr.md
```diff
@@ -1,8 +1,56 @@
...
+* Do you test so specifically that WebMock would be the better tool?
+* Does your project have special needs for tweaks: https://blog.arkency.com/3-tips-to-tune-your-vcr-in-tests/
+* How often do you want to re-record your cassettes? Out-of-date replays may give you a false sense of safety.
```
sihu, line 54: * how easy is it to reproduce the cases? (e.g. if it's not the happy case or it's very complicated to re-record, resp. to bring it to that state, it might be better to mock the method and fake the status code/error code)

ruby_on_rails/vcr.md
```diff
@@ -1,8 +1,56 @@
...
+* Where do you want to re-record your cassettes? Maybe nightly on the CI?
```
sihu, line 55: did we ever do/use that?

sihu (approved):

ruby_on_rails/vcr.md
```diff
@@ -1,8 +1,56 @@
...
-Here's an article on about some tweaks which can be made to a VCR setup: https://blog.arkency.com/3-tips-to-tune-your-vcr-in-tests/
+Here's an example configuration we use at Renuo:
+
+```rb
+# spec/rails_helper.rb
+
+RSpec.configure do |config|
+  # …
+
+  # Disable VCR completely for tests that are not tagged with :vcr
```
coorasse, line 14: This is the main point that I am not sure about, why don't you want to be warned that a remote call is executed? Isn't there a risk that your tests depend on external services?

ruby_on_rails/vcr.md
```diff
@@ -1,8 +1,56 @@
...
+  c.default_cassette_options = {
+    decode_compressed_response: true,
+    allow_unused_http_interactions: false,
+    record: ENV['VCR'] ? ENV['VCR'].to_sym : :once, # re-record with VCR=all
+    drop_unused_requests: true # only when re-recording
+  }
+
+  # Filter out sensitive data from the cassettes
+  env_keys = YAML.load_file('config/application.example.yml').filter do |k, v|
+    (!k.in? %w[test production development]) && (!v.in? %w[false true])
```
coorasse, line 44: also simple numbers can be a big issue, I'd exclude also values that are just numbers

coorasse (commented):
I believe we could have warnings about cassettes that are too old, but at the same time, a custom web mock response might be very old as well 🤔 .
I am not sure about the assumption where model tests shouldn't use VCR. I do this all the time. 
I like that VCR warns you if you if you perform a remote call and did not use either VCR or Webmock...but this comes by default, so I am not sure that I understand how this improves the situation. I believe that I need to look into it closer together.

ruby_on_rails/vcr.md
schmijos: No, your assumption is wrong. _WebMock_ catches everything. Also if your test is not marked with `:vcr`.

ruby_on_rails/vcr.md
schmijos: That's a good point. The current configuration is maybe not precise enough. We would allow HTTP calls to ourselves.
Be aware though that for example request specs are not actual HTTP requests.

ruby_on_rails/vcr.md
schmijos: Even though you're technically correct, I disagree. I could not remember anything longer than `VCR`. For example I like very much that @sihu uses `DEBUG=true` for `selenium_chrome` driver instead of headless. It's very unspecific but very rememberable.

ruby_on_rails/vcr.md
schmijos: Yes. What's a good heuristic for secrets? Longer than 7 characters?

PR #339, opened 2023-08-23
Update capybara.rb

schmijos:
selenium-webdriver 4.11 brought breaking headless config updates for the cucumber Capybara configuration in one of our projects. While this did not explicitly break, I revisited this config anyways and split it up.


coorasse:
I am not understanding what issue is this solving. I don't have this file in fredi and it seems to work well.

schmijos:
> I am not understanding what issue is this solving. I don't have this file in fredi and it seems to work well.

I need a proper Capybara configuration for cucumber.

PR #341, opened 2023-10-24
Use bash from env for better portability

Luz:
Hi
According to [Stackoverflow](https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang#10383546) it looks like the portability of the scripts could be improved by having this new style for the bash shebang.

### DIFF
--- a/bin/check
+++ b/bin/check
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 set -euo pipefail
 
--- a/bin/setup
+++ b/bin/setup
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -euo pipefail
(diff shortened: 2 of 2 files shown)


rnestler (approved):
Detected the NixOS user :wink: 

LGTM, but the Mac people should maybe also chime in?

coorasse:
What issue are we trying to solve? Is there an existing problem?

rnestler:
> What issue are we trying to solve? Is there an existing problem?

Read the linked stackoverflow: As is the scripts won't run on operating systems that put bash in a different place than `/bin/bash` (or don't symlink `/bin/bash` to the place where bash resides). Examples seem to be OpenBSD, FreeBSD, NixOS

coorasse (approved):
All right! On mac it works in both ways.

PR #342, opened 2023-11-30
Enable branch coverage

hunchr:
TICKET-21097

### DIFF
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -38,8 +38,9 @@ You should know exactly why you are adding each one of them and why is necessary
     add_filter 'app/mailers/application_mailer.rb'
     add_filter 'app/models/application_record.rb'
     add_filter '.semaphore-cache'
+    enable_coverage :branch
+    minimum_coverage line: 100, branch: 100
   end
-  SimpleCov.minimum_coverage 100
   ```
 
   to run code coverage and exclude files with less then 5 lines of code.
--- a/templates/spec/spec_helper.rb
+++ b/templates/spec/spec_helper.rb
@@ -8,8 +8,10 @@ unless ENV['NO_COVERAGE']
     add_filter 'app/jobs/application_job.rb'
     add_filter 'app/mailers/application_mailer.rb'
     add_filter 'app/models/application_record.rb'
+    add_filter '.semaphore-cache'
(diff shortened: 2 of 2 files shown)


rnestler (approved):

coorasse (changes requested):
I think we should set a minimum coverage to 100, as per line coverage. Teams can always decide to reduce it in their project or to exclude branches. Enabling without enforcing brings no value. It will simply be ignored.

hunchr:
> I think we should set a minimum coverage to 100, as per line coverage. Teams can always decide to reduce it in their project or to exclude branches. Enabling without enforcing brings no value. It will simply be ignored.

the coverage was already set to 100% before
```rb
SimpleCov.minimum_coverage 100
```

coorasse (commented):
Can you please explain to me what `primary_coverage` does? I expect a `SimpleCov.minimum_branch_coverage` option since they are two separate stats.

ruby_on_rails/rspec.md
```diff
@@ -38,6 +38,8 @@ You should know exactly why you are adding each one of them and why is necessary
     add_filter 'app/mailers/application_mailer.rb'
     add_filter 'app/models/application_record.rb'
     add_filter '.semaphore-cache'
+    enable_coverage :branch
+    primary_coverage :branch
   end
   SimpleCov.minimum_coverage 100
```
rnestler, line 44:
```suggestion
  SimpleCov.minimum_coverage line: 100, branch: 100
```

This is how you configure branch coverage.

rnestler (changes requested):

ruby_on_rails/rspec.md
hunchr: Interesting, this is never mentioned in the simplecov docs

hunchr:
> Can you please explain to me what `primary_coverage` does? I expect a `SimpleCov.minimum_branch_coverage` option since they are two separate stats.

I disabled `primary_coverage` again and used @rnestler's suggestion

ruby_on_rails/rspec.md
coorasse: https://github.com/simplecov-ruby/simplecov#minimum-coverage

coorasse (approved):
This looks good now!

PR #346, opened 2024-01-12
Add appsignal docs

firemind:


coorasse (commented):
good. we need to link it at some point of the setup

firemind:
> good. we need to link it at some point of the setup

Added it to the rails readme and made sentry optional since I assume we want to default to appsignal for error monitoring

coorasse (approved):

coorasse:
Thanks!

PR #347, opened 2024-01-24
Add info on how to fix severity on appsignal

firemind:
This came after quite some testing by @rnestler and me.

The reason we require such a heavy solution is that we need to display two custom pieces of information (`severity` and `request_id`) and we need to display both of them in `logfmt` format.

Rails comes with `ActiveSupport::TaggedLogging` which gets information about the `request_id` injected by rack middleware. That info then gets passed to the formatter. Unfortunately it is limited displaying tags in the `[TAG]` format. We could re-implement all of `ActiveSupport::TaggedLogging` and the `ActiveSupport::TaggedLogging::Formatter` to do the same thing but format it as `logfmt`. However, since there will never be both of them active at the same time, overriding it seems like the better solution.

This information is global to the request. The severity on the other hand can differ from each call to the logger so it can't just be set as a tag.

### DIFF
--- a/ruby_on_rails/appsignal.md
+++ b/ruby_on_rails/appsignal.md
@@ -34,6 +34,46 @@ Message format: **logfmt**
 
 Then add this log drain using the heroku commands displayed.
 
+## Correct Severity
+
+According to the docs, getting the severity to be anything but "INFO" is not possible using the heroku drain.
+
+However, there is now a way to send the `"severity=XYZ"` logfmt information and have that be applied correctly in appsignal. Unfortunately, just setting this seems to break the recognition of `request_id` in the format `[1234-5678]`. So we have to override the `ActiveSupport::TaggedLogging::Formatter` to add both the `severity` and the `request_id` in logfmt syntax.
+
+```
+module ActiveSupport
+  module TaggedLogging
+    module Formatter
+      def call(severity, timestamp, progname, msg)
+        super(severity, timestamp, progname, "severity=#{severity} #{tags_text}#{msg}")
+      end
+
+      def tags_text
+        tags = current_tags
+        tags.inject('') do |text, tag|
+          if tag.is_a? Hash
+            text += tag.map { |k,v| "#{k}=#{v} " }.join
+          else
+            raise 'we use our own tags!'
+          end
+        end
+      end
+    end
+  end
+end
+```
+
+and
+
+```
+# config/environments/production.rb
+Rails.application.configure do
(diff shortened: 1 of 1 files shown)


ruby_on_rails/appsignal.md
```diff
@@ -34,6 +34,46 @@ Message format: **logfmt**
 
 Then add this log drain using the heroku commands displayed.
 
+## Correct Severity
+
+According to the docs, getting the severity to be anything but "INFO" is not possible using the heroku drain.
```
rnestler, line 39: Can you link to these docs here?

ruby_on_rails/appsignal.md
```diff
@@ -34,6 +34,46 @@ Message format: **logfmt**
...
+
+However, there is now a way to send the `"severity=XYZ"` logfmt information and have that be applied correctly in appsignal. Unfortunately, just setting this seems to break the recognition of `request_id` in the format `[1234-5678]`. So we have to override the `ActiveSupport::TaggedLogging::Formatter` to add both the `severity` and the `request_id` in logfmt syntax.
```
rnestler, line 41: In the herold project I saw that you attempted to revert back to the less invasive logger overide? So this didn't work out I guess? https://github.com/renuo/herold-planner/commit/9f2ab8eeceeacd0e2975f89e3ca2f73239462183

ruby_on_rails/appsignal.md
```diff
@@ -34,6 +34,46 @@ Message format: **logfmt**
...
+  end
+end
+```
+
+and
+
+```
+# config/environments/production.rb
+Rails.application.configure do
+  config.log_tags = [lambda { |request| {request_id: request.request_id} }]
```
rnestler, line 71: Maybe mention in a comment, that this is our special log tag format with key and values.

coorasse (approved):
I am fine with this but I'd like some references on Appsignal and maybe they should also mention something in their docs?

ruby_on_rails/appsignal.md
firemind: No, it doesn't parse the `request_id` anymore using that setup.

ruby_on_rails/appsignal.md
firemind: done

ruby_on_rails/appsignal.md
firemind: added

schmijos (approved):
I like that you put it as an opt-in.

PR #348, opened 2024-01-24
Minimal docs for fastlane

schmijos:


fastlane.md
```diff
@@ -0,0 +1,19 @@
+# fastlane
+
+We manage our iOS app certificates with [fastlane](https://fastlane.tools/).
+They are managed in this private repository: [fastlane-ios-certificates](https://github.com/renuo/fastlane-ios-certificates).
```
lukasbischof, line 4: We also have a separate one for Mode Weber since we have separate certificates there: https://github.com/renuo/fastlane-ios-modeweber-certificates

lukasbischof (approved):

PR #350, opened 2024-02-19
rails: Use bash with hardened options in convenience scripts

rnestler:


coorasse (approved):
I guess is a linux thing? Anyway, I tested it and it works on mac too

PR #351, opened 2024-02-19
rails: Use ENV.fetch instead of ENV[] || default

rnestler:
When generating a new project with Rails 7.1 it uses `ENV.fetch("RAILS_LOG_LEVEL", "info")`, so we can just change the string to `"warn"` there.

### DIFF
--- a/ruby_on_rails/app_initialisation.md
+++ b/ruby_on_rails/app_initialisation.md
@@ -109,7 +109,7 @@ create the initializer for figaro in `config/initializers/figaro.rb`:
 
   ```ruby
   config.force_ssl = true # uncomment
-  config.log_level = ENV['RAILS_LOG_LEVEL']&.to_sym || :warn # change
+  config.log_level = ENV.fetch("RAILS_LOG_LEVEL", "warn") # change
   ```
 
 * Update `config/environments/development.rb` settings:


ruby_on_rails/app_initialisation.md
```diff
@@ -109,7 +109,7 @@ create the initializer for figaro in `config/initializers/figaro.rb`:
 
   ```ruby
   config.force_ssl = true # uncomment
-  config.log_level = ENV['RAILS_LOG_LEVEL']&.to_sym || :warn # change
+  config.log_level = ENV.fetch("RAILS_LOG_LEVEL", "warn") # change
```
firemind, line 112: Did you mean to change this from a symbol to a string? It's unlikely to matter but the official docs do use symbols: https://guides.rubyonrails.org/debugging_rails_applications.html#log-levels

rnestler: Well the generated code when creating a new rails 7 project had `ENV.fetch("RAILS_LOG_LEVEL", "info")`

firemind: Ok, in that case the docs are probably outdated.

firemind (approved):

coorasse (approved):

PR #354, opened 2024-02-27
depfu: Mention engine updates and automatic merging

rnestler:


ruby_on_rails/depfu.md
```diff
@@ -10,3 +10,49 @@ repository.
...
+
+```yaml
+# .github/workflows/depfu_autoapprove.yml
+name: Depfu auto-approve
+on:
+  pull_request_target:
+    types: [opened]
+
+permissions:
+  contents: write
```
lukasbischof, line 45: Is the `contents: write` permission actually needed? A quick look at the table at https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs suggests that the pull-requests permission might be the only requirement

rnestler: Good point. I'm not sure how I ended up with specifying both originally.

rnestler: Tested it and it works. Fixed.

lukasbischof (approved):

PR #360, opened 2024-03-13
Replace pry with irb

hunchr:
todo: explain why

### DIFF
--- a/ruby_on_rails/suggested_libraries.md
+++ b/ruby_on_rails/suggested_libraries.md
@@ -20,9 +20,7 @@ end
 group :development, :test do
   gem 'awesome_print'
   gem 'bullet'
-  gem 'pry-byebug'
-  gem 'pry-doc'
-  gem 'pry-rails'
+  gem 'debug'
 end
 
 group :production do


coorasse (commented):
Thank you @hunchr !
Two things:
* terminal: I believe irb catched up with pry by now. I am in favor of removing pry.
* debug: I cannot judge. I use rubymine. Therefore I am open to any solution wished by people using debug.

I'd like at least a couple of more approvals on this one.

rnestler (approved):
I haven't tried it too much with debugging yet to compare. A few notes:

 * `irb` is the standard, so I prefer using it.
 * The less dependencies the better
 * `pry-rails` should improve the `bin/rails console`, but I actually like the `irb` provided one way more.

schmijos:
I would drop all of them. I use `binding.irb` now.

rnestler:
> I would drop all of them. I use binding.irb now.

But for that you'll need to require the irb gem, no?

hunchr:
> But for that you'll need to require the irb gem, no?

the irb gem is already included by rails
image

ruby_on_rails/suggested_libraries.md
```diff
@@ -20,9 +20,7 @@ end
 group :development, :test do
   gem 'awesome_print'
   gem 'bullet'
-  gem 'pry-byebug'
-  gem 'pry-doc'
-  gem 'pry-rails'
+  gem 'debug'
```
rnestler, line 23:
```suggestion
```

rnestler (approved):
Let's remove it and use `binding.irb`

schmijos:
> > I would drop all of them. I use binding.irb now.
> 
> But for that you'll need to require the irb gem, no?

No. It's a default gem. You can require it directy: https://stdgems.org/irb/

rnestler:
> If you mean bundling it: No. It's a default gem. You can require it directly:

I wanted to avoid needeing to do `require 'irb'; binding.irb`, but it's not needed since it seems in scope anyways.

schmijos (approved):

PR #361, opened 2024-03-22
Remove `bin/fastcheck` from `bin/check` script

lukasbischof:
Executing `bin/fastcheck` in `bin/check` causes it to be executed twice on the CI and thus increasing CI time. In some projects, this was already removed. We should do the same for all other projects as well to keep it consistent and fast

### DIFF
--- a/ruby_on_rails/app_initialisation.md
+++ b/ruby_on_rails/app_initialisation.md
@@ -58,7 +58,7 @@ They are always idempotent (runnable multiple times).
 * Add a `bin/check` file. It will run all the automated tests. It's mainly used in our CI.
 
   ```sh
-  echo "#\!/usr/bin/env bash\nset -euo pipefail\nbin/fastcheck\nbin/rails zeitwerk:check" > bin/check
+  echo "#\!/usr/bin/env bash\n\nset -euo pipefail\n\nbin/rails zeitwerk:check" > bin/check
   ```
 
 * Make the new scripts executable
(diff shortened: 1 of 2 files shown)


ruby_on_rails/app_initialisation.md
```diff
@@ -58,7 +58,7 @@ They are always idempotent (runnable multiple times).
 * Add a `bin/check` file. It will run all the automated tests. It's mainly used in our CI.
 
   ```sh
-  echo "#\!/usr/bin/env bash\nset -euo pipefail\nbin/fastcheck\nbin/rails zeitwerk:check" > bin/check
+  echo "#\!/usr/bin/env bash\nset -euo pipefail\n\nbin/rails zeitwerk:check" > bin/check
```
lukasbischof, line 61: I think a double new line increases readability and we also have it this way in other projects as well (e.g. mtextur, remms)

rnestler (approved):
Totally agree. In a related note: I'd rename it from "fastcheck" to "lint", since in practice we only run linters in it.

lukasbischof:
> I'd rename it from "fastcheck" to "lint", since in practice we only run linters in it.

I also like this naming better than fastcheck (in my opinion, fastcheck is a bad name as it's not apparent what it does, especially in presence of a check script. The name suggest that it just runs the tests faster). However, this would either produce different namings in newer projects or a lot of pull requests moving a file

coorasse (approved):

coorasse:
Agree! I am also in favor of renaming, but let's move on with this change in the meantime.

PR #362, opened 2024-03-27
Add rails_helper I18n locale reset configuration

CuddlyBunion341:
Does it make sense to move the rspec configuration inside of the I18n templates instead of having it in the rspec section?
https://github.com/renuo/applications-setup-guide/blob/main/i18n.md

### DIFF
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -94,6 +94,12 @@ You should know exactly why you are adding each one of them and why is necessary
       config.before(:each, type: :system, js: true) do
         driven_by ENV['SELENIUM_DRIVER']&.to_sym || :selenium_chrome_headless
       end
+
+      config.around do |example|
+        locale_before = I18n.locale
+        example.call
+        raise "This test did not reset the locale properly" if I18n.locale != locale_before
+      end
     end
 
   # config/application.example.yml
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -95,10 +95,8 @@ You should know exactly why you are adding each one of them and why is necessary
         driven_by ENV['SELENIUM_DRIVER']&.to_sym || :selenium_chrome_headless
       end
 
-      config.around do |example|
-        locale_before = I18n.locale
-        example.call
-        raise "This test did not reset the locale properly" if I18n.locale != locale_before
+      config.before do |example|
+        I18n.locale = I18n.default_locale
       end
     end
 


ruby_on_rails/rspec.md
```diff
@@ -94,6 +94,12 @@ You should know exactly why you are adding each one of them and why is necessary
       config.before(:each, type: :system, js: true) do
         driven_by ENV['SELENIUM_DRIVER']&.to_sym || :selenium_chrome_headless
       end
+
+      config.around do |example|
+        locale_before = I18n.locale
+        example.call
+        raise "This test did not reset the locale properly" if I18n.locale != locale_before
```
coorasse, line 101: Why does it matter? You reset it anyway before each test. 
I have often tests where I want to test stuff on a different locale, I don't want to be forced to reset it at the end of the test.
Maybe `I18n.with_locale { example.call }` would be a better alternative?

coorasse (commented):
I also have the locale reset before each test in cobra/fredi. But I'd not raise at the end, otherwise wrap

ruby_on_rails/rspec.md
schmijos: Agreed. I like @coorasse's suggestion better.

CuddlyBunion341: > You reset it anyway before each test.
> @coorasse 
Shouldn't this be enforced? It seems like having the snippet it would aid more than it harms.

The suggestion to use with_locale sounds better but this would be need defined inside of each spec that uses a different locale, no?
```rb
it "tests something" do
  I18n.with_locale(:de) do
    # do something
  end
end
```
Or for a describe block:

```rb
describe MyClass do
	around(:each) do |example|
	  I18n.with_locale(:de) do
	    example.run
	  end
	end
	
	it "does something using the german locale" do
	  # do something
	end
end
```

Instead of manually setting and restoring the locale.

By using the code snippet in the rails_helper we could prevent developers from forgetting to reset the locale. (If they choose to not use the I18n around example).

Or should it be possible to configure the locale how we configure vcr?

```rb
it "does something using the french locale", locale: :fr do
  # do something
end
```

What are your thoughts @schmijos?

CuddlyBunion341: I do quite like this:
```rb
around(:each) { |example| I18n.with_locale(:de) { example.run } }
```

coorasse: `around(:each) { |example| I18n.with_locale(I18n.default_locale) { example.run } }`

Yes, I think we should have a good default to avoid leaking and random failure. I am fine with both solutions:
* around
* setting it to default locale before each test

People can then find their way of setting different locales in their apps if needed.

rnestler: What is wrong with the following?
```ruby
# spec/rails_helper.rb
config.before do |example|
  I18n.locale = I18n.default_locale
end
```

So every test starts with the default locale.

CuddlyBunion341: This is fine by me.

ruby_on_rails/rspec.md
```diff
@@ -94,6 +94,12 @@ You should know exactly why you are adding each one of them and why is necessary
...
+      end
```
CuddlyBunion341, line 102:
```suggestion
      config.before do |example|
        I18n.locale = I18n.default_locale
      end
```

schmijos (approved):

coorasse:
The line was already there: https://github.com/renuo/applications-setup-guide/pull/362/files#diff-d7277f8c823d1e6eefd0329793c31cbd408ded80fcfeff439cda456da9619febR82

PR #364, opened 2024-04-05
Update fastcheck to include a diaeresis check

coorasse:
These characters are not supported by all fonts and might cause issues on PDF.

### DIFF
--- a/templates/bin/fastcheck
+++ b/templates/bin/fastcheck
@@ -30,3 +30,10 @@ coffeelint -f .coffeelint.json app/javascripts/**/*.coffee
 
 # for javascript
 yarn eslint app/javascripts spec/javascripts
+
+# for diaeresis chracters
+output=$(find . -name '*.yml' -exec grep -nH "̈" {} +); echo "$output"; [ -z "$output" ]
+if [ $? -ne 0 ]; then
+  echo "There are combining diaeresis characters in the YML files. Remove them."
+  exit 1
+fi
--- a/templates/bin/fastcheck
+++ b/templates/bin/fastcheck
@@ -31,7 +31,7 @@ coffeelint -f .coffeelint.json app/javascripts/**/*.coffee
 # for javascript
 yarn eslint app/javascripts spec/javascripts
 
-# for diaeresis chracters
+# for diaeresis characters
 output=$(find . -name '*.yml' -exec grep -nH "̈" {} +); echo "$output"; [ -z "$output" ]
 if [ $? -ne 0 ]; then
   echo "There are combining diaeresis characters in the YML files. Remove them."


hunchr (commented):
some languages like german need ä/ö/ü so i don't see why this check would be useful

edmunteanu:
> some languages like german need ä/ö/ü so i don't see why this check would be useful

It's useful because there are two ways to write ä / ö / ü, one with a / o / u + diaeresis character ( ¨ ) or the ä / ö / ü directly. It can be that some fonts don't support the a / o / u + ¨, but support the ä / ö / ü.

This was an issue on cobra for example, where the font didn't know what to do with the ¨ and displayed it as an artefact:
CleanShot 2024-04-09 at 13 37 08@2x

coorasse:
Yes, this PR misses some context. Thanks @edmunteanu

templates/bin/fastcheck
```diff
@@ -30,3 +30,10 @@ coffeelint -f .coffeelint.json app/javascripts/**/*.coffee
 
 # for javascript
 yarn eslint app/javascripts spec/javascripts
+
+# for diaeresis characters
+output=$(find . -name '*.yml' -exec grep -nH "̈" {} +); echo "$output"; [ -z "$output" ]
+if [ $? -ne 0 ]; then
```
rnestler, line 36:
```suggestion
if find . -name '*.yml' -exec grep -nH "̈" {} + ; then
```

No point for the complicated dance with storing the output and checking if it is empty, if we just want to print the matches and error out if there are any.

rnestler: It was also pointed out by [shellcheck](https://www.shellcheck.net/) as a code smell:
```
In templates/bin/fastcheck line 36:
if [ $? -ne 0 ]; then
     ^-- SC2181 (style): Check exit code directly with e.g. 'if ! mycmd;', not indirectly with $?.
```

coorasse: I think we then need to switch the check:

```
➜  find . -name '*.yml' -exec grep -nH "̈" {} +
./config/locales/pdf.de-CH.yml:15:         Wir freuen uns, Ihnen hiermit einen temporären Punktepass zu überreichen.
./config/locales/mailers.de-CH.yml:81:      intro_html: Vielen Dank für Ihr Interesse an der Gratis Manor Treuekarte. Um
➜  echo $?                                    
0
```

```
➜  find . -name '*.yml' -exec grep -nH "̈" {} +
➜  echo $?                                    
1
```

because it's 0 if it finds something

rnestler: Ah sorry yeah. Should have tested.

rnestler:
```
if find . -name '*.yml' -exec grep -nH "<0308>" {} +; then echo '!!!found!!!'; fi
./config/locales/mailers.de-CH.yml:75:        enjoy_html: Wir wünschen Ihnen viel Spass beim Shopping und Punktesammeln.
./config/locales/mailers.de-CH.yml:81:      intro_html: Vielen Dank für Ihr Interesse an der Gratis Manor Treuekarte. Um
./config/locales/pdf.de-CH.yml:15:         Wir freuen uns, Ihnen hiermit einen temporären Punktepass zu überreichen.
./config/locales/pdf.de-CH.yml:16:         Mit dem Punktepass können Sie von Rabatten profitieren, zeitlich befristet Manor Punkte sammeln und einsetzen,
./config/locales/pdf.de-CH.yml:17:         jedoch nicht Ihre Einkäufe bezahlen.
./config/locales/pdf.de-CH.yml:19:         damit Sie jederzeit von allen Vorteilen profitieren können.\n\n
./config/locales/pdf.de-CH.yml:20:         Der Punktepass ist während 14 Tagen ab Ausstellungsdatum gültig und wird in allen Manor Filialen akzeptiert.
./config/locales/pdf.de-CH.yml:21:         Bewahren Sie ihn sorgfältig auf.\n\n
./config/locales/pdf.de-CH.yml:22:         Wir wünschen Ihnen viel Vergnügen beim Einkaufen in unseren Warenhäusern."
./config/locales/pdf.de-CH.yml:44:         Die Manor Sofortkarte ist während 14 Tagen ab Ausstellungsdatum gültig und wird in allen Manor Warenhäusern akzeptiert.
./config/locales/mailers.fr-CH.yml:75:        enjoy_html: Wir wünschen Ihnen viel Spass beim Shopping und Punktesammeln.
!!!found!!!
```

sihu (approved):
I would have liked to know what problem you are trying to solve. as far as i understood PDFs do not support such characters. I'm happy to add this test if you think it's helpful. i think it's rather an edgecase that effects not a lot of our projects.

schmijos (commented):
I'm not sure if this test is consequently good. While it may be good enough to distinguish between umlaut and a/o/u dieresis there are actual dieresis like in [naïve](https://en.wikipedia.org/wiki/%C3%8F).

coorasse:
Yes, there are corner cases of corner cases.

coorasse:
I'll drop this from the ASG. I don't think there's a consensus.

PR #365, opened 2024-04-10
Add explicit hint to update plugin libraries

edmunteanu:
Since the ASG recommends installing ruby with asdf (and I don't know of a way to tell asdf to automatically update its libraries periodically), I think it's worth having a small hint to update them manually before installing the latest ruby version. That way one can be sure they're really getting the latest version.

### DIFF
--- a/ruby_on_rails/app_initialisation.md
+++ b/ruby_on_rails/app_initialisation.md
@@ -2,7 +2,7 @@
 
 ## Default Rails setup
 
-* Ensure that your asdf plugins are up to date with `asdf plugin update -all`.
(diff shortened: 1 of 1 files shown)


ruby_on_rails/app_initialisation.md
```diff
@@ -2,9 +2,11 @@
 
 ## Default Rails setup
 
+* Ensure that your asdf plugins are up to date with `asdf plugin update -all`.
```
lukasbischof, line 5:
```suggestion
* Ensure that your asdf plugins are up to date with `asdf plugin update --all`.
```

PR #367, opened 2024-05-22
Cleanup lograge config

simon-isler:


schmijos (approved):
I like where this goes. Either initializer only for production or put it directly in `production.rb`.

PR #371, opened 2024-07-10
Update .sass-lint.yml

coorasse:
As discussed during the Rails Bootstrap UI Workshop with @ekyburz and @ignaciosy


ignaciosy (approved):
thanks!

On the same topic, I haven't really worked on projects including this linter, and it now felt VERY restrictive and opinionated...it hindered me more than help.

coorasse:
I also don't have cit in my projects. I cannot really give you a generic opinion .

PR #375, opened 2024-09-04
Update the first green test section to use /up rails endpoint

coorasse:
This PR simplifies the initial setup by removing the need of writing a home controller. It makes use of the default HealthController provided by Rails and provides an example of how to extend and customize it.


.mdlrc
```diff
@@ -1,2 +1,2 @@
-rules "~MD013", "~MD026", "~MD029", "~MD032", "~MD036", "~MD002", "~MD034", "~MD031", "~MD028"
+rules "~MD013", "~MD026", "~MD029", "~MD032", "~MD036", "~MD002", "~MD034", "~MD031", "~MD028", "~MD033", "~MD027"
```
coorasse, line 1: I disagree with `MD033`. I want to use HTML. in this case `<br>` tags. MD027 was instead giving a false positive.

ruby_on_rails/rspec.md
```diff
@@ -116,20 +116,21 @@ We are now going to write a first test to ensure that the whole configuration is
...
 * `bin/check` should be green :white_check_mark:
 * Write the test [`spec/system/health_spec.rb`](../templates/spec/system/health_spec.rb)
-* Write the controller [`app/controllers/home_controller.rb`](../templates/app/controllers/home_controller.rb)
-* Add `get 'home/check'` to `config/routes.rb`
-
 * Run `bin/check` and the test should pass and coverage is 100%.
 
 Commit and push your changes! :tada:
 
+> ⭐️ `/up` is the defualt Health check path for Rails. [Read about it in the guides](https://edgeapi.rubyonrails.org/classes/Rails/HealthController.html).<br>
```
schmijos, line 123:
```suggestion
> ⭐️ `/up` is the default Health check path for Rails. [Read about it in the guides](https://edgeapi.rubyonrails.org/classes/Rails/HealthController.html).<br>
```

.mdlrc
schmijos: There are two other options:
* Using a backslash `\` at the end of the line 
* Using two spaces `  ` at the end of the line

See 
CleanShot 2024-09-04 at 13 48 31@2x
https://daringfireball.net/projects/markdown/syntax#block

schmijos (approved):

ruby_on_rails/app_initialisation.md
```diff
@@ -15,18 +15,19 @@ You can do this with `gem update rails`. Beware of beta versions.
...
-> If you do not need a DB you may rethink the fact that you may not need Rails at all :) Take a look at [Sinatra](http://www.sinatrarb.com/) or [Angular](https://angular.io/)
+> ⚠️ You may want to choose a different database than Postgres, but most of the time this will be your choice.\
+> If you do not need a DB you may rethink the fact that you may not need Rails at all: Take a look at [Sinatra](http://www.sinatrarb.com/) or [Angular](https://angular.io/)\
 > You might also need actionmailbox of course, so always double-check the parameters that you are using.
-> Check if you've got a `.ruby-version` file. If not, create one and enter the Ruby version.
 
-> ⭐️ This setup does not include either js-bundling nor css-bundling by default. It will start with the simplest possible Rails setup and will use sprockets and importmaps.
-> If you need to do fancy stuff, discuss with your team the opportunity of including a js-bundling and css-bundling tool.
+> ⭐️ This setup does not include either js-bundling nor css-bundling by default.\
+> It will start with the simplest possible Rails setup and will use sprockets and importmaps.\
```
simon-isler, line 23:
```suggestion
> It will start with the simplest possible Rails setup and will use sprockets and importmaps.\
```

can we go for `propshaft`?

simon-isler: sprockets is not the simplest possible Rails setup

coorasse: not yet. I don't have a solution to compile scss yet. I have some other moving pieces that I have to figure out. propshaft will be default in rails8

simon-isler: @coorasse what about `cssbundling-rails`? why don't you want to got with `propshaft` yet?

PR #377, opened 2024-09-04
Update erb-linter template

coorasse:


templates/.erb-lint.yml
```diff
@@ -1,20 +1,26 @@
...
       Layout/InitialIndentation:
         Enabled: false
+      Layout/LineLength:
+        Enabled: true
       Layout/TrailingEmptyLines:
         Enabled: false
       Layout/TrailingWhitespace:
         Enabled: false
+      Layout/BlockAlignment:
+        Enabled: false
```
coorasse, line 23: this has false-positives

CuddlyBunion341: Could you please provide an example?
This seems to be a pretty important rule:
https://docs.rubocop.org/rubocop/cops_layout.html#layoutblockalignment

CuddlyBunion341 (approved):

templates/.erb-lint.yml
```diff
@@ -1,20 +1,26 @@
 ---
-glob: "**/app/views/**/*.{html,text,js}{+*,}.erb"
 EnableDefaultLinters: true
+exclude:
+  - '**/rails_admin/*'
+  - "vendor/**/*"
 linters:
+  SpaceInHtmlTag:
+    enabled: false
```
simon-isler, line 8:
```suggestion
    enabled: false
    NoJavascriptTagHelper:
    	enabled: true
	correction_style: 'plain'
```

https://github.com/Shopify/erb_lint/tree/main?tab=readme-ov-file#nojavascripttaghelper

simon-isler (commented):
nice, could you also link the template in the markdown? I think I was missing this

templates/.erb-lint.yml
coorasse: I cannot write:
CleanShot 2024-09-04 at 17 22 27@2x

I'd have to write:
CleanShot 2024-09-04 at 17 22 49@2x

templates/.erb-lint.yml
coorasse: I am not sure that I understand this correctly.

simon-isler: the indentation is a bit off. I'd add the `NoJavascriptTagHelper` with `correction_style: 'plain'`. the default adds  //<![CDATA[ and //]]> markers (see https://github.com/Shopify/erb_lint/tree/main?tab=readme-ov-file#nojavascripttaghelper)

simon-isler:
> nice, could you also link the template in the markdown? I think I was missing this

👀

PR #379, opened 2024-09-08
Gemfile reads .ruby-version

coorasse:
This is a change we apply to all our projects. There's no need to do that manually as long as people know why is needed. That's why I introduce also a list of questions at the end of the template.

### DIFF
--- a/ruby_on_rails/template.rb
+++ b/ruby_on_rails/template.rb
@@ -39,7 +39,7 @@ after_bundle do
 
   puts "We want to be sure that you are aware of what has been done by the Renuo template."
   puts "Please read the following list of changes and confirm that you understand them."
-  ask "The Gemfile ruby version has been set to the version in the .ruby-version file. You know what this means.  Ok?"
+  ask "The Gemfile ruby version has been set to the version in the .ruby-version file. You know what this means. Ok?"
   ask "Your project is now using renuocop instead of rubocop-rails-omakase. You know both gems and why it has been replaced. Ok?"
   ask "A .rubocop.yml file has been created with the default configuration. You know what this means. Ok?"
   ask "A bin/fastcheck script has been created to run rubocop with the default configuration."
--- a/ruby_on_rails/template.rb
+++ b/ruby_on_rails/template.rb
@@ -43,5 +43,5 @@ after_bundle do
   ask "Your project is now using renuocop instead of rubocop-rails-omakase. You know both gems and why it has been replaced. Ok?"
   ask "A .rubocop.yml file has been created with the default configuration. You know what this means. Ok?"
   ask "A bin/fastcheck script has been created to run rubocop with the default configuration."
-  ask "You can run the script with `bin/fastcheck`. Youm know why we have a bin/fastcheck file. Ok?"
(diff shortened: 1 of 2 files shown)


ruby_on_rails/template.rb
```diff
@@ -2,6 +2,13 @@
 
 gsub_file "Gemfile", /gem "rubocop-rails-omakase"/, "gem \"renuocop\""
 
+# add "ruby file: ".ruby-version" to the Gemfile under the line starting with "source"
+insert_into_file "Gemfile", after: /^source.*\n/ do
+  <<~RUBY
+    ruby File.read(".ruby-version").strip
```
CuddlyBunion341, line 8: Finally!

ruby_on_rails/template.rb
```diff
@@ -30,3 +33,15 @@
 end
 
 run "chmod +x bin/fastcheck"
+
+after_bundle do
```
CuddlyBunion341, line 37: I like the way you prompt the CLI users to understand how the CLI is setting up the project. Thanks!

ruby_on_rails/template.rb
```diff
@@ -10,10 +17,6 @@
   RUBOCOP
 end
 
-after_bundle do
```
CuddlyBunion341, line 13: Why is this not necessary anymore?

CuddlyBunion341 (approved):
Looks good to me 👍🏻 ! Thanks for taking care of this stuff @coorasse!

ruby_on_rails/template.rb
simon-isler, line 8:
```suggestion
    ruby file: ".ruby-version"
```

ruby_on_rails/template.rb
```diff
@@ -30,3 +33,15 @@
 end
 
 run "chmod +x bin/fastcheck"
+
+after_bundle do
+  run "bundle exec rubocop -A"
+
+  puts "We want to be sure that you are aware of what has been done by the Renuo template."
+  puts "Please read the following list of changes and confirm that you understand them."
+  ask "The Gemfile ruby version has been set to the version in the .ruby-version file. You know what this means.  Ok?"
```
simon-isler, line 42:
```suggestion
  ask "The Gemfile ruby version has been set to the version in the .ruby-version file. You know what this means. Ok?"
```

ruby_on_rails/template.rb
```diff
@@ -30,3 +33,15 @@
...
+  ask "Your project is now using renuocop instead of rubocop-rails-omakase. You know both gems and why it has been replaced. Ok?"
+  ask "A .rubocop.yml file has been created with the default configuration. You know what this means. Ok?"
+  ask "A bin/fastcheck script has been created to run rubocop with the default configuration."
+  ask "You can run the script with `bin/fastcheck`. Youm know why we have a bin/fastcheck file. Ok?"
```
simon-isler, line 46:
```suggestion
  ask "You can run the script with `bin/fastcheck`. You know why we have a bin/fastcheck file. Ok?"
```

ruby_on_rails/template.rb
coorasse: Right!

PR #381, opened 2024-09-25
Only not risky autofixes

schmijos:
Risky autofixes shouldn't be needed, I guess.

### DIFF
--- a/ruby_on_rails/template.rb
+++ b/ruby_on_rails/template.rb
@@ -54,7 +54,7 @@ end
 run "chmod +x bin/fastcheck"
 
 after_bundle do
-  run "bundle exec rubocop -A"
+  run "bundle exec rubocop -a"
 
   puts "We want to be sure that you are aware of what has been done by the Renuo template."
   puts "Please read the following list of changes and confirm that you understand them."


schmijos:
But they have a very limited ruleset. We may break things because we use way more rules.

schmijos:
And it's not default anymore: https://github.com/rails/rails/pull/51782/files

coorasse:
I think this can be entirely removed anyway. Rails does it.

coorasse (commented):
To me is either we don't do anything, or I am ok with the `-A`. I have enough code coverage everywhere to catch bugs. I'd rather do that than manually adjust rubocop offenses. But if more people think otherwise I can live with that.

coorasse:
I added some more people, so that we can come to a conclusion on this PR.

rnestler:
> I have enough code coverage everywhere to catch bugs. 

It's not about your code though but about generated code for a new project, which may subtly fail.

rnestler (approved):
I very much agree with the comment in https://github.com/rails/rails/pull/53138#issuecomment-2386392714 that if you want to automatically apply unsafe cops you can just mark them as safe.

So running rubocop automatically should be done with `-a`.

firemind (approved):
To my understanding this is a one-off run of rubocop right after initializing a new, empty project. It seems very unlikely to have unsafe things to fix, but if it somehow does, I agree it would be better not to fix them at that point.

ignaciosy (approved):
I agree with Mike in that, being a one-off when creating a new project, I'd rather rubocop not break anything, as there are many initial-setup things that could already be misconfigured.
Afterwards, if `bin/check` goes for the risky fixes I wouldn't mind because they'd be easier to identify

schmijos:
Let me conclude:

* We disagree with the change here.
* But we would agree with the change in https://github.com/rails/rails/pull/53138

So I'm closing this as well.

PR #388, opened 2024-09-30
Migrate from hub to gh command

CuddlyBunion341:


coorasse (commented):
Please describe why this change

CuddlyBunion341:
On second thought, the 'hub' command seems quite useful due to its more concise usage. It would be simpler if we didn't have to install gh and hub for the application setup, but I'll leave it be.

PR #394, opened 2025-01-17
ruby-on-rails: Skip kamal by default when generating a new project

rnestler:
Very likely we won't deploy using Kamal in almost all cases.


coorasse (approved):
Agree.

PR #395, opened 2025-01-29
Improve setup and explanation of renuo fetch-secrets

coorasse:
TICKET-23362

Adds an explanation on how to use the 1Password vault and the necessary steps to set it up.

### DIFF
--- a/ruby_on_rails/app_initialisation.md
+++ b/ruby_on_rails/app_initialisation.md
@@ -94,6 +94,17 @@ create the initializer for figaro in `config/initializers/figaro.rb`:
 
 * Run `bin/setup` again.
 
+### Secrets
+
+We store the secrets necessary to configure the project locally in a 1password Item.
+Create a new Item for the project called `[project-name] project secrets` of type note.
+Right click on the vault and select `Copy Private link`.
+
+* Run the command `renuo fetch-secrets --init [the vault private link]` to create an empty secrets file.
+
+The Item contains the fields that represent the ENV variables. You can use Text or Password fields.
+Check existing projects for an example of the usage.
+
 #### Configuration customisation
 
 * Update `config/application.rb` and set the default language and timezone
--- a/ruby_on_rails/template.rb
+++ b/ruby_on_rails/template.rb
@@ -2,6 +2,23 @@
 
 gsub_file "Gemfile", /gem "rubocop-rails-omakase"/, "gem \"renuocop\""
 
+insert_into_file "Gemfile", after: /^group :development do\n/ do
+  <<~RUBY
+    gem "renuo-cli", require: false
+  RUBY
+end
+
+# replace bin/rails db:prepare with bin/rails db:setup in bin/setup
+gsub_file "bin/setup", "bin/rails db:prepare", "bin/rails db:setup"
+
+# add the renuo fetch-secrets command in bin/setup, before bin/rails db:setup
+insert_into_file "bin/setup", before: "\n  puts \"\\n== Preparing database ==\"" do
+  <<-RUBY
+  puts "\\n== Fetching 1password dependencies =="
+  system! 'renuo fetch-secrets'  
+  RUBY
+end
+
 # add "ruby file: ".ruby-version" to the Gemfile under the line starting with "source"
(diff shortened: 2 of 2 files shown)


ruby_on_rails/app_initialisation.md
```diff
@@ -94,6 +94,17 @@ create the initializer for figaro in `config/initializers/figaro.rb`:
 
 * Run `bin/setup` again.
 
+### Secrets
+
+We store the secrets necessary to configure the project locally in a 1password vault.
+Create a new vault for the project called `[project-name] project secrets` of type note.
+Right click on the vault and select `Copy Private link`.
+
+* Run the command `renuo fetch-secrets --init [the vault private link]` to create an empty secrets file.
```
coorasse, line 103: https://github.com/renuo/renuo-cli/pull/165

ruby_on_rails/app_initialisation.md
```diff
@@ -94,6 +94,17 @@ create the initializer for figaro in `config/initializers/figaro.rb`:
...
+
+The Vault contains the Text fields that represent the ENV variables.
+Check existing projects for an example of the usage.
+
 #### Configuration customisation
```
simon-isler, line 108:
```suggestion
### Configuration customisation
```

ruby_on_rails/app_initialisation.md
```diff
@@ -94,6 +94,17 @@ create the initializer for figaro in `config/initializers/figaro.rb`:
 
 * Run `bin/setup` again.
 
+### Secrets
+
+We store the secrets necessary to configure the project locally in a 1password vault.
+Create a new vault for the project called `[project-name] project secrets` of type note.
+Right click on the vault and select `Copy Private link`.
```
simon-isler, line 101: instead of a new vault called `[project-name] project secrets`, I think it would be nice to have a vault per project. @schmijos and I made this an operations goal for this year. See https://docs.google.com/document/d/1kqig4ITDiSciXu6mGtIg8b0bUsqQXjNl8De6gu5Y3Hg/edit?tab=t.0#heading=h.tr0aaqqpaniu

@coorasse what do you think about this?

ruby_on_rails/app_initialisation.md
```diff
@@ -94,6 +94,17 @@ create the initializer for figaro in `config/initializers/figaro.rb`:
...
+
+* Run the command `renuo fetch-secrets --init [the vault private link]` to create an empty secrets file.
+
+The Vault contains the Text fields that represent the ENV variables.
+Check existing projects for an example of the usage.
```
simon-isler, line 106: To me, this is an item and not a vault? Vaults are used to organize items, whereas items store the secrets, isn't it?

ruby_on_rails/template.rb
```diff
@@ -2,6 +2,23 @@
 
 gsub_file "Gemfile", /gem "rubocop-rails-omakase"/, "gem \"renuocop\""
 
+insert_into_file "Gemfile", after: /^group :development do\n/ do
+  <<~RUBY
+    gem "renuo-cli", require: false
+  RUBY
+end
```
simon-isler, line 9: nice, thanks for implementing it 👍🏼

ruby_on_rails/app_initialisation.md
hunchr: > contains the Text fields

i tried it with password fields and it works too. maybe that's better for sensitive env vars?

ruby_on_rails/app_initialisation.md
coorasse: I don't see the need to have a separate vault per customer. 

But when we will have, we can of course change this part here as well 👍

simon-isler (approved):

lukasbischof:
@coorasse @simon-isler 
When `renuo-cli` is added to the development dependencies, how do we ensure that we use the latest version of the CLI? The current version depends on `activesupport ~> 8.0`, making it incompatible with older projects. For example, when trying to add the dependency to `andres-geotechnik` (which is a Rails 7.2 project) the latest available version of the CLI is 1.6.

PR #396, opened 2025-01-30
Adapt Ruby code snippets to new renuocop rules

rnestler:
Can we run `rubocop` with `renuocop` rules on the application setup guide snippets somehow?


coorasse (approved):
Thanks!

PR #398, opened 2025-02-26
Add yarn note

shindi-renuo:
Not every project has a necessity for `yarn`.

This PR adds a little note where `bin/yarn` gets introduced.

### DIFF
--- a/ruby_on_rails/linting_and_automatic_check.md
+++ b/ruby_on_rails/linting_and_automatic_check.md
@@ -34,6 +34,8 @@ end
 
 ## SCSS lint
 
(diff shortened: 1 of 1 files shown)


ruby_on_rails/linting_and_automatic_check.md
```diff
@@ -34,6 +34,8 @@ end
 
 ## SCSS lint
 
+> _Note_: You may not need `yarn` in your Project. Please consult with your teammates.
```
coorasse, line 37: I think we should not run the command with `bin/yarn` but just `npm` and write a note that the Semaphore configuration might need to be adjusted

coorasse (approved):

PR #407, opened 2025-05-01
Do not start server on bin/setup

coorasse:
`bin/setup` should simple setup the app not also start the server. The rest of the applications setup guide does not expect the server to be started.

See also https://github.com/rails/rails/commit/7ed09d3a387a780278700ac6edac3e97aa41aa3a

### DIFF
--- a/ruby_on_rails/template.rb
+++ b/ruby_on_rails/template.rb
@@ -19,6 +19,9 @@ insert_into_file "bin/setup", before: "\n  puts \"\\n== Preparing database ==\""
   RUBY
 end
 
+# remove the block unless ARGV.include?("--skip-server"). We don't want to start the server directly during setup.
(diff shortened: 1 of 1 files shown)


ruby_on_rails/template.rb
```diff
@@ -19,6 +19,23 @@
...
+  RUBY
+end
+
+# add the renuo-cli gem to the Gemfile under the line starting with "gem \"rails\""
+insert_into_file "Gemfile", after: /^gem \"rails\"\n/ do
+  <<~RUBY
+    gem "renuo-cli", require: false
+  RUBY
+end
+
```
simon-isler, line 38: duplicate?

schmijos (changes requested):
Please check the duplicate.

simon-isler (approved):

schmijos (approved):

PR #408, opened 2025-05-12
Add ruleset configuration

CuddlyBunion341:
TICKET-23953

[Rulesets](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets) are a new feature of GitHub. They are the preffered way of defining protection rules compared to the classic options.

The new Rulesets offer more flexibility and control compared to the "Classic branch protection rules".
We propose that we should embrace the new standard for upcoming repositories.

### DIFF
--- a/configure_git_repository.md
+++ b/configure_git_repository.md
@@ -19,6 +19,7 @@ Please stick to it unless you have special needs.
   * `develop`
     * Enforcement status: `Active`
     * Branch targeting criteria: `develop`
+    * Bypass list: add `Repository Admin` Role with *allow for pull requests only* option
     * Restrict deletions
     * Require linear history
     * Require a pull request before merging
--- a/configure_git_repository.md
+++ b/configure_git_repository.md
@@ -1,4 +1,4 @@
-# Configure the Git Repository
+# Configure the GitHub Repository
 
 These are the suggested configurations for our GitHub repositories.
 Please stick to it unless you have special needs.
@@ -9,15 +9,26 @@ Please stick to it unless you have special needs.
     * Disable *Allow merge commits* and *Allow rebase merging*
     * Allow auto-merge
     * Automatically delete head branches
+    * Always suggest updating pull request branches
 * Manage access
   * Add *staff* team as a collaborator with Admin access
   * Add *security* team as collaborator with Write access
 * Branches
   * Default branch: either `main` or `develop` depending on whether you want one or two environments.
-  * Add these rules for the two branches `develop` and `main`:
-    * Require pull request reviews before merging
-    * Require status checks to pass before merging (after you configured the CI add it to the required checks)
-    * Always suggest updating pull request branches
+* Rules/Rulesets
+  * `develop`
+    * Enforcement status: `Active`
+    * Branch targeting criteria: `develop`
+    * Restrict deletions
+    * Require linear history
+    * Require a pull request before merging
+    * Require status checks to pass
+        * Select `ci/semaphore/push`
+    * Block force pushes
+  * `main` (same as develop but...)
+    * Branch targeting criteria: `main`
+    * ❌ Require a pull request before merging
+    * ❌ Require status checks to pass
 
 * Autolink references
   * Add a new Autolink reference with:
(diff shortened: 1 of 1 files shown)


coorasse (commented):
Two questions:
* is this an improvement in any direction? the new rulesets offer more flexibility, but if we don't use it, does it mean more complexity for us and we would be perfectly fine with the existing branch protection?
* we were trying to fix the problem that we can push to develop and we don't want that, while allowing everyone to still bypass checks and merge if they want. Is this now tackled?

hunchr:
> does it mean more complexity for us

it's very similar to the classic branch protection. it's not more complex.

> we were trying to fix the problem that we can push to develop and we don't want that, while allowing everyone to still bypass checks and merge if they want. Is this now tackled?

yes, you can uncheck `Require status checks to pass` if you want to merge a failing pr

coorasse:
> yes, you can uncheck Require status checks to pass if you want to merge a failing pr
We need to sit together and you show me how it behaves. I want to always be able to bypass checks, acknowledge the risk, and click the merge button. If I read this correclty, if my CI fails, or I did not get a review, I need to go and change branch protection rules, and this is not what we want at Renuo.

coorasse (commented):
as per my last comment, show me a demo next time we are in the office

configure_git_repository.md
```diff
@@ -17,6 +17,7 @@ Please stick to it unless you have special needs.
   * `develop`
     * Enforcement status: `Active`
     * Branch targeting criteria: `develop`
+    * Bypass list: Add `Repository Admin` Role with "allow For pull requests only" option
```
hunchr, line 20: we could also just leave this unchecked
image

hunchr: otherwise, we bypass everything, no?

coorasse:
@CuddlyBunion341 I am not sure anymore that I am ok in blocking pushes to develop. Example below: I really don't want to go through a branch and a PR for this:
CleanShot 2025-05-22 at 09 35 32@2x

configure_git_repository.md
coorasse: There's a big difference: by allowing to "bypass" you have to explicitly check a checkbox where you acknowledge you are breaking the rules and you need to press a "semi-red" button. This makes you really conscious of your choice.

This is an important bit I asked Dani to add, because otherwise we are not free to choose anymore and the workflow is enforced. That's not how I want people to work at Renuo: I want to give a workflow to follow and a set of established and automated rules. Everyone has the option to take responsibility of their actions and break those rules.

cc @schmijos

rnestler:
> I don't really see how and why they should be "preferred". There's no deprecation from github about branch protection. So this first point does not really convince me.

The UI of GitHub really pushes you in the direction of rulesets, which makes it awkward when configuring for the first time, since it seems like one is using a deprecated "classic" functionality:

image

schmijos:
Consider also that Github pushes you to use fine grained personal access tokens. That's part of the compliance circus. Maybe Github will be done for sooner than we anticipate.

CuddlyBunion341:
> @CuddlyBunion341 I am not sure anymore that I am ok in blocking pushes to develop. Example below: I really don't want to go through a branch and a PR for this: CleanShot 2025-05-22 at 09 35 32@2x

That's a good point @coorasse.
The problem with pushes to develop is that if you block force pushes, then undoing the push to develop is a clumbersome as you need to disable the ruleset temporarily.

coorasse:
I am happy to switch, but I don't want to give up my freedom, so the new ruleset should not block me from pushing directly to develop if I really want to. Also, rules should be bypassable by administrators as of today.

So I recap my points here:

* I would like to keep the possibility to push on develop
* We can also remove the "disable force push" rule. I am not that interested.
* I still want administrators to be able to bypass rules (so everyone)
* I really don't care what color the button on github is 

Given the above we can switch to the new branch protection rule 👍

hunchr:
> We can also remove the "disable force push" rule. I am not that interested.

ok, i removed it

> I still want administrators to be able to bypass rules (so everyone):

that's still possible https://github.com/renuo/applications-setup-guide/blob/9f2c5c9925604bdbc4158914ed70cfb39409338d/configure_git_repository.md?plain=1#L20

schmijos (approved):
I assume we tested the behavior of these settings.

PR #409, opened 2025-05-26
Recommend parallel_tests

coorasse:
parallel_tests allows us to make use of all our processors instead of a single one. It's definitely time to have it by default.

What's next:
* [ ] make rspec optional (or opt-in) and suggest a possible experimentation of minitest
* [ ] suggest splitting of system tests from everything else from the beginning

### DIFF
--- a/ruby_on_rails/rspec.md
+++ b/ruby_on_rails/rspec.md
@@ -1,6 +1,6 @@
 # Setup RSpec
 
-Even though Rails uses Minitest per default, RSpec is the *de-facto* standard at Renuo. We love RSpec and we strongly suggest to use it.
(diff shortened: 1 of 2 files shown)


ruby_on_rails/rspec.md
```diff
@@ -61,51 +62,64 @@ You should know exactly why you are adding each one of them and why is necessary
   Please check the [spec_helper template](../templates/spec/spec_helper.rb)
 
 * Add the configurations:
```
coorasse, line 64: only indentation changes.

simon-isler (approved):
Looking forward to try it out

I agree with the changes that you mentioned in the description

PR #417, opened 2025-07-28
Update configure_git_repository.md

ddbrendan:
Actually, I think this entire step might be redundant.
When you configure Semaphore CI (following the "[Configure the CI](https://open-source.renuo.ch/applications-setup-guide/ruby_on_rails/configure_ci.html)" section of the guide), it automatically:

Integrates with GitHub
Creates the status check ci/semaphoreci/push: [project-name]
Reports the check status on every PR

You don't need to manually add any status check requirements in the GitHub branch protection rules because it's already handled by the Semaphore integration.

Even without the configuration, the Semaphore check still happens:
image

Sorry for being unclear!

### DIFF
--- a/configure_git_repository.md
+++ b/configure_git_repository.md
@@ -23,9 +23,6 @@ Please stick to it unless you have special needs.
     * Restrict deletions
     * Require linear history
     * Require a pull request before merging
-    * Require status checks to pass
-      * Select `ci/semaphore/push`
-      * **Note: This status check name may be outdated
     * Block force pushes
   * `main` (same as develop but...)
     * Branch targeting criteria: `main`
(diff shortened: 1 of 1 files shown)


CuddlyBunion341 (changes requested):
I don't quite follow this change..
Can you please describe what name may be outdated and why in the PR?

configure_git_repository.md
```diff
@@ -25,6 +25,7 @@ Please stick to it unless you have special needs.
     * Require a pull request before merging
     * Require status checks to pass
       * Select `ci/semaphore/push`
+      * **Note: This status check name may be outdated
```
CuddlyBunion341, line 28: So we can skip this part?
```suggestion
```

CuddlyBunion341 (approved):

PR #441, opened 2026-04-07
Add hint regarding Deploio Economy database

sislr:
TICKET-5516

see also https://github.com/renuo/renuo-cli/pull/291


coorasse (approved):
Why is it not straightforward?

sislr:
> Why is it not straightforward?

It would require us to create a new DB from scratch and migrate all the data. Nine will work on providing a smooth transition

PR #443, opened 2026-04-24
Link to VPS setup guide

firemind:


sislr (commented):
I'd link it once we have the initial VPS setup documented as well. Let's keep it as a draft for now

PR #445, opened 2026-04-29
Replace emoji shortcodes with unicode emoji and drop emojify step

cb341:
The mdbook deploy and lint workflows previously ran shonfeder/emojitsu to expand GitHub-style :shortcode: emoji into unicode at build time. Replace the shortcodes with unicode directly in the source so the build step is no longer needed, and drop the Emojify Book steps from both workflows.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -15,12 +15,6 @@ jobs:
         uses: peaceiris/actions-mdbook@v2
         with:
           mdbook-version: 'latest'
-      - name: Emojify Book
-        run: |
-          set -x
-          wget --no-verbose https://github.com/shonfeder/emojitsu/releases/download/0.1.1/gh-actions-emojitsu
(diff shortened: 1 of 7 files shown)


.github/workflows/gh-pages.yml
```diff
@@ -14,13 +14,7 @@ jobs:
...
         uses: peaceiris/actions-mdbook@v2
         with:
-          mdbook-version: 'latest'
-      - name: Emojify Book
-        run: |
-          set -x
-          wget --no-verbose https://github.com/shonfeder/emojitsu/releases/download/0.1.1/gh-actions-emojitsu
-          chmod +x gh-actions-emojitsu
-          find . -type f -name "*.md" -exec ./gh-actions-emojitsu emojify -i {} \;
+          mdbook-version: '0.4.40'
```
cb341: why pin?

coorasse (approved):

PR #447, opened 2026-04-29
Add mdbook preprocessor to rewrite template links

cb341:
Previously file links in the mdbook docs would simply download the file. This can be annoying because it takes an extra step to get to the source code of the file if it needs to be opened first.

My proposal is to simply rewrite the file links such that they point onto our GitHub instead.

**BEFORE**

https://github.com/user-attachments/assets/0cff633a-ccbd-468e-9617-8304bc7ab61d

**AFTER**

https://github.com/user-attachments/assets/91a64912-20a9-4019-89d7-bf860c317ee6

### DIFF
--- a/bin/mdbook-rewrite-links
+++ b/bin/mdbook-rewrite-links
@@ -8,6 +8,12 @@
 
 require "json"
 
+# mdbook invokes preprocessors with `supports <renderer>` to probe compatibility.
+# Exit 0 to signal support; no stdin payload is sent in that mode.
+if ARGV[0] == "supports"
+  exit 0
+end
+
 GITHUB_BLOB = "https://github.com/renuo/application-setup-guide/blob/main"
 
 # Matches ](path) where path points into templates/ and does NOT end in .md.
@@ -28,10 +34,6 @@ def walk(section)
     chapter["content"] = rewrite(chapter["content"]) if chapter["content"]
     chapter["sub_items"]&.each { |item| walk(item) }
   end
-end
-
-if ARGV[0] == "supports"
-  exit 0
 end
 
 _context, book = JSON.parse($stdin.read)
--- a/bin/mdbook-rewrite-links
+++ b/bin/mdbook-rewrite-links
@@ -1,0 +1,40 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# mdbook preprocessor: rewrite relative links to templates/* (non-.md) into
+# absolute GitHub blob URLs so the published book points to GitHub instead of
+# offering a raw-file download. Source markdown keeps relative paths so local
+# editing and raw-GitHub viewing still work.
+
+require "json"
+
+GITHUB_BLOB = "https://github.com/renuo/application-setup-guide/blob/main"
+
+# Matches ](path) where path points into templates/ and does NOT end in .md.
+# Handles both ../templates/... and templates/... forms.
+LINK_RE = %r{\]\((\.\.?/)?(templates/[^)\s]+?)\)}
+
+def rewrite(markdown)
+  markdown.gsub(LINK_RE) do
+    path = Regexp.last_match(2)
+    next Regexp.last_match(0) if path.end_with?(".md")
+    "](#{GITHUB_BLOB}/#{path})"
+  end
+end
+
+def walk(section)
+  if section.is_a?(Hash) && section["Chapter"]
+    chapter = section["Chapter"]
+    chapter["content"] = rewrite(chapter["content"]) if chapter["content"]
+    chapter["sub_items"]&.each { |item| walk(item) }
+  end
+end
+
+if ARGV[0] == "supports"
+  exit 0
+end
+
+_context, book = JSON.parse($stdin.read)
+items = book["items"] || book["sections"] || []
+items.each { |section| walk(section) }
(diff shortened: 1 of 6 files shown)


ruby_on_rails/cucumber.md
```diff
@@ -41,8 +41,7 @@ If you add Cucumber to an existing project, test a real page instead of using th
 
 Add the following files:
 
-* [`app/controllers/home_controller.rb`](../templates/app/controllers/home_controller.rb) (If you haven't done so
-already) in the [RSpec](rspec.md) section.
+* [`app/controllers/home_controller.rb`](../templates/app/controllers/home_controller.rb)
```
cb341:
```suggestion
* [`app/controllers/home_controller.rb`](../templates/app/controllers/home_controller.rb) (If you haven't done so already) in the [RSpec](rspec.md) section.
```

templates/app/controllers/home_controller.rb
cb341, line 1: This file didn't exist for some reason despite being referenced int he docs

bin/mdbook-rewrite-links
```diff
@@ -0,0 +1,66 @@
...
+exit 0 if ARGV == ["supports", "html"]
+
+LINK_RE = %r{(\[[^\]]*\])\((\.\.?/)?(templates/[^)\s]+?)\)}
+
+LANGS = {
+  ".rb" => "ruby", ".js" => "js", ".ts" => "ts", ".sh" => "sh",
+  ".yml" => "yaml", ".yaml" => "yaml", ".json" => "json",
+  ".scss" => "scss", ".css" => "css", ".html" => "html",
+  ".feature" => "gherkin", ".erb" => "erb", ".jsx" => "jsx"
+}
```
cb341: This is debatable.

We could either choose the easy route and serve it as RAW text.
OR we could take this approach and create virtual sections where the file contents are wrapped in a code block.

bin/mdbook-rewrite-links
cb341, line 1: As this is mostly a behavioral change for mdbook itslef and doesn't affect the actual contents.

bin/mdbook-rewrite-links
cb341: The upside of the code block is the copy button and the more consistent look and feel.

bin/mdbook-rewrite-links
```diff
@@ -0,0 +1,66 @@
...
+      inlined[path] ||= File.read(file)
+      "#{label}(#{prefix}#{path}.html)"
+    end
+  end
+  chapter["sub_items"]&.each(&walk)
+end
+items.each(&walk)
+
+inlined.each do |path, content|
+  fence = "`" * ([3, (content.scan(/`+/).map(&:length).max || 0) + 1].max)
```
cb341: ?

hunchr:
I have no idea what this is about. Please assign someone who knows the ASG.

templates/app/controllers/home_controller.rb
hunchr: Is this still needed? Rails has `/up`.

bin/mdbook-rewrite-links
```diff
@@ -0,0 +1,40 @@
...
+
+def walk(section)
+  if section.is_a?(Hash) && section["Chapter"]
+    chapter = section["Chapter"]
+    chapter["content"] = rewrite(chapter["content"]) if chapter["content"]
+    chapter["sub_items"]&.each { |item| walk(item) }
+  end
+end
+
+if ARGV[0] == "supports"
```
hunchr: What is this?

templates/app/controllers/home_controller.rb
cb341: that's fair..
but then we need a different boilerplate setup for the initial cucumber test.

https://open-source.renuo.ch/application-setup-guide/ruby_on_rails/cucumber.html?highlight=cucum#cucumber

hunchr (changes requested):
Why don't we just link to GH directly?

bin/mdbook-rewrite-links
cb341: No idea, let's try dropping it
```suggestion
```

templates/app/controllers/home_controller.rb
hunchr: Good point, but
> If you add Cucumber to an existing project, test a real page instead of using the given example test.

I'm assuming there is already at least one controller. Otherwise, adding Cucumber before being able to test anything doesn't make sense.

bin/mdbook-rewrite-links
cb341: "supports" arg is used to check for preprocessor compatability by mdbook:

https://rust-lang.github.io/mdBook/for_developers/preprocessors.html

```rs
fn make_app() -> Command {
    Command::new("nop-preprocessor")
        .about("A mdbook preprocessor which does precisely nothing")
        .subcommand(
            Command::new("supports")
                .arg(Arg::new("renderer").required(true))
                .about("Check whether a renderer is supported by this preprocessor"),
        )
}
```

cb341:
> Why don't we just link to GH directly?

That works poorly when you want to use the docs in an offline environment.
With relative paths you can still navigate files.

hunchr (approved):

PR #452, opened 2026-07-03
Update appsignal config to respect deploio git revision

sislr:
https://docs.appsignal.com/ruby/integrations/rails#configure-appsignal-in-an-initializer
> We recommend using our config/appsignal.rb config file or environment variables to configure AppSignal. By default, it’s not possible to configure AppSignal from a Rails initializer, because AppSignal loads before your application’s initializers, in order to be able to report errors that take place during initialization.


coorasse (commented):
why is this not in the initializers folder anymore?

sislr:
> why is this not in the initializers folder anymore?

Well, please read the PR description

